PR #1, opened 2015-01-27
Update version to 0.2.1 in bower.json

hahahana:
Hi, thanks for creating this mixin! It worked perfectly for my needs. Just FYI, you mentioned in the readme that this was available on Bower, but it actually wasn't, so I went ahead and registered it for you. I hope that was okay, I just really needed to use it ASAP.


Pomax:
totally forgot to add that change. thanks!

PR #4, opened 2015-02-13
clean up handlers to avoid leaking memory

jergason:
Since every component that uses this mixin will add to the module state when it mounts, we end up leaking memory. We Remove the components and event handlers from their respective arrays when the components unmount to avoid this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -68,7 +68,11 @@
       var pos = registeredComponents.indexOf(this);
       if( pos>-1) {
         var fn = handlers[pos];
+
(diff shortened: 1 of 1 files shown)


Pomax:
if you move those splice instructions down a little to inside where know `fn` actually exists, I'll be happy to merge this in.

jergason:
@Pomax the deed is done

Pomax:
superwin

Pomax:
0.2.2 released

Issue #6, opened 2015-02-21
DOMNode may not exist


isaaczafuta:
If a component using this mixin returns null in the initial render, `this.getDOMNode()` will return null in componentDidMount. If the component later renders something more substantial, clicks that are inside of its new DOM node won't be registered as 'inside'.

Pomax:
Hmm. `render()`s entire purpose is to do just that: render the master container, possibly with content but not necessarily so, so even if you have nothing to show yet, your `getInitialState` paired with the `this.props` values your instance is created with means `render()` can always generate a thing. Having it yield a null is somewhat an indicator that React is being used incorrectly, and I'd rather not patch the code to silently accept anti-patterns. As such, much as I appreciate the PR, I'm going to close it, since I think it's more important that people follow the correct React methodologies.

isaaczafuta:
Thanks for the feedback.

I think I slightly misstated the issue here-- The fact that I focused on the 'initial' render and returning null was a poor choice of example. Perhaps consider something like a dynamic image loader, with a `render` along the lines of:
```javascript
render: function () {
    if (this.state.imageLoaded) {
        return (<img src={this.props.src} />);
    } else {
        return (<span>nope, still loading</span>);
    }
}
```
When `imageLoaded` changes, the dom node that was captured in `componentDidMount` will become stale. While the `img` and `span` could be wrapped to mitigate this issue, this conditional return pattern seems pretty prevalent, and manages to save a node in the dom.

Pomax:
That still feels weird, since you're not generating a stable UI element for the component this way. I'd strongly recommend at the very least using:

```
render function() {
  var content = <span>Loading...</span>;
  if(this.state.imageLoaded) {
    content = <img src={this.props.src} />
  }
  return <div className="fft-img-23">{content}</div>;
}
```

at which point the event handler works correctly.

PR #7, opened 2015-02-21
handle case where initial render returns null

isaaczafuta:
Address issue #6 by getting the dom node in the event handler.


Pomax:
See #6 for the rationale behind closing this PR rather than merging it in.

PR #8, opened 2015-03-12
Fix comment-doc for required component function

davidtheclark:
Unless that comment is getting at something I misunderstood, the method that is required is `handleClickOutside`, not `onClickOutside`, right?


Pomax:
correct. nice catch

Issue #9, opened 2015-03-20
Ignore clicks inside certain elements


arunthampi:
I have a requirement where I need the mixin to callback if its click outside a component but with a few exceptions. The method I used was to add the class `ignore-react-onclickoutside` to whichever elements that I want the mixin to ignore and change this line

https://github.com/Pomax/react-onclickoutside/blob/b9842bf81f75ed7fcebab38f503eb52cea9b9927/index.js#L51

to 

```javascript
found = (source === localNode || source.className.match(/ignore-react-onclickoutside/));
```

If you think this is a valid feature request (and solution), I can send a pull request (with the appropriate documentation changes as well).

Pomax:
If you're up for writing a PR for it, I'll be happy to look it over - it sounds like a good additional control to have!

arunthampi:
Done

Pomax:
thanks! left two comments on how we can do this without needing regexp matching, other than that this looks extremely sensible to me!

Pomax:
v0.2.3 published to npm

arunthampi:
sweet, thanks for the super quick turnaround!

PR #10, opened 2015-03-20
Ignore elements

arunthampi:
In reference to https://github.com/Pomax/react-onclickoutside/issues/9

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,6 +33,8 @@
   var registeredComponents = [];
   var handlers = [];
 
+  var IGNORE_CLASS = 'ignore-react-onclickoutside';
+
   return {
     componentDidMount: function() {
       if(!this.handleClickOutside)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -33,6 +33,8 @@
   var registeredComponents = [];
   var handlers = [];
 
+  var IGNORE_CLASS_REGEXP = new RegExp('ignore-react-onclickoutside');
```
Pomax: no need for this to be a regexp, actually, let's just make this `var IGNORE_CLASS = "ignore-react-onclickoutside";`, see below

index.js
```diff
@@ -48,7 +50,7 @@
           // thinking in terms of Dom node nesting, running counter
           // to React's "you shouldn't care about the DOM" philosophy.
           while(source.parentNode) {
-            found = (source === localNode);
+            found = (source === localNode || source.className.match(IGNORE_CLASS_REGEXP));
```
Pomax: classList lets us as targeted questions, so we can do this:
```
source.classList.contains(IGNORE_CLASS)
```
without needing to match regexp on the HTML node "class" node attribute values.

arunthampi:
Thanks for the feedback, this is much nicer -- I've squashed the fixing commit

Pomax:
nice! merging, and I'll spin an updated version

PR #11, opened 2015-04-03
Check the top most node to see if it is inside or has class

cymen:
There is a bug when iterating upwards -- if we stop checking when the `source.parentNode` is undefined, we miss the last `source`. At least for the `IGNORE_CLASS`, this is a bug. I think it's a bug for the regular check too.

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,7 +49,7 @@
           // a layered approach, too, but that requires going back to
           // thinking in terms of Dom node nesting, running counter
           // to React's "you shouldn't care about the DOM" philosophy.
-          while(source.parentNode) {
+          while(source) {
             found = (source === localNode || source.classList.contains(IGNORE_CLASS));
             if(found) return;
             source = source.parentNode;


Pomax:
the last source is going to be <html>, we really don't need to check that =)

cymen:
I ran into issues where `.parentNode` was not undefined so we never got to `<html>`. I think do to some wacky things in a plugin (how it was creating DOM). The last node we did get to was the one node I could actually insert a class on from outside of the plugin. But it makes sense to fix that plugin issue so I'll go look at that.

Pomax:
Do you happen to have some code for that case? It would be interesting to see. This will fail if there is not top `<html>` like in a DOMFragment, but as those are not part of the active DOM, can't receive user clicks anyway (only 'faked' events). (basically this plugin is for test live nodes, since it assumes you're working in React, rather than manipulating DOM nodes yourself)

PR #12, opened 2015-05-03
Add support for IE < 10 and other browsers without classList

Daniel15:
Adds support for browsers without classList support, primarily Internet Explorer 8 and 9. Also added support for legacy IE event model for IE 8.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,6 +28,13 @@
 }(this, function () {
   "use strict";
 
+  // Use classList if available, else fall back to manual string comparison
+  var hasClass = document.body.classList
+    ? function(el, className) { return el.classList.contains(className); } 
+    : function(el, className) {
+      return (' ' + el.className + ' ').indexOf(' ' + className + ' ') > -1;
+    }
+
   // Use a parallel array because we can't use
   // objects as keys, they get toString-coerced
   var registeredComponents = [];
@@ -42,7 +49,7 @@
 
       var fn = (function(localNode, eventHandler) {
         return function(evt) {
-          var source = evt.target;
+          var source = evt.target || evt.srcElement;
           var found = false;
           // If source=local then this event came from "somewhere"
           // inside and should be ignored. We could handle this with
@@ -50,7 +57,7 @@
           // thinking in terms of Dom node nesting, running counter
           // to React's "you shouldn't care about the DOM" philosophy.
           while(source.parentNode) {
-            found = (source === localNode || source.classList.contains(IGNORE_CLASS));
+            found = (source === localNode || hasClass(source, IGNORE_CLASS));
             if(found) return;
             source = source.parentNode;
           }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -28,6 +28,13 @@
 }(this, function () {
   "use strict";
 
+  // Use classList if available, else fall back to manual string comparison
+  var hasClass = document.body.classList
+    ? function(el, className) { return el.classList.contains(className); } 
+    : function(el, className) {
+      return (' ' + el.className + ' ').indexOf(' ' + className + ' ') > -1;
```
Pomax: this would be more robust as `return el.className.split(" "').indexOf(classname) > -1`

index.js
```diff
@@ -58,8 +65,15 @@
...
 
-      document.addEventListener("mousedown", fn);
-      document.addEventListener("touchstart", fn);
+      if (document.addEventListener) {
+        // W3C events
+        document.addEventListener("mousedown", fn);
+        document.addEventListener("touchstart", fn);
+      } else if (document.attachEvent) {
+        // Old IE
+        document.attachEvent("onmousedown", fn);
```
Pomax: I'm not particularly interested in supporting IE8 or below, which are the only versions of IE where attachEvent is necessary. If you are on one of these versions of IE, you are also on an operating system that isn't even supported by Microsoft anymore, so in this case I will not accommodate those setups. They'll just have to get an OS that's less than 15 years old and abandoned by the company that used to support it.

Pomax:
I'm conflicted about this PR. On the one hand, it's nice to offer support for old browsers, but in this case it's for browsers that you're only going to be on if (a) you're on windows 98 or XP, which cannot run IE 9 or above, but are also no longer supported by Microsoft, or (b) you're on windows vista, 7 or 8.x, in which case you can be on IE11 right now, and still being on IE9 is your own obstinate choice. Either way, I'd much rather confront you with the fact that you're part of the problem than add shims that make it seem like it's okay to keep using these browsers. It's not.

index.js
Daniel15: Support for classList is 89% globally: http://caniuse.com/#search=classList. While high, that's still 1 in 10 people that use browsers that don't support it.

Daniel15:
That's fair enough. I'll just keep this version in my own branch and use that whenever I want IE8 support. I don't normally care about old IE any more, but this was the only tweak I needed (other than loading es5-shim) to make a site work properly in it, so I felt it wasn't a very big overhead.

It does look like classList has better support now, last I checked that wasn't the case so I'm glad browsers have caught up.

index.js
Pomax: but what's the spread? I trust that value, but don't trust that it matters for people who can run React apps at all.

Pomax:
For future reference, http://caniuse.com is pretty useful to get the overview tables for most features like this.

PR #14, opened 2015-05-18
Internet Explorer support

ken210:
Internet Explorer [lack Element.classList](http://caniuse.com/#feat=classlist) full support.

So i just add the [classlist-polyfill](https://www.npmjs.com/package/classlist-polyfill) as dependency.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -25,5 +25,8 @@
     "bower_components",
     "test",
     "tests"
-  ]
+  ],
+  "dependencies": {
+    "classlist-polyfill": "https://github.com/yola/classlist-polyfill.git"
+  }
 }
--- a/index.js
+++ b/index.js
@@ -30,6 +30,7 @@
 
   // Use a parallel array because we can't use
   // objects as keys, they get toString-coerced
+  require('classlist-polyfill');
   var registeredComponents = [];
   var handlers = [];
 
--- a/package.json
+++ b/package.json
@@ -20,5 +20,8 @@
   },
   "bugs": {
     "url": "https://github.com/Pomax/react-onclickoutside/issues"
+  },
+  "dependencies": {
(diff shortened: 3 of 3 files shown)


Pomax:
Actually still-supported versions of Internet Explore [definitely support those](http://caniuse.com/#search=classList), so the PR fixes the mixin for browsers that are no longer supported, for operating systems that Microsoft no longer even supports While I'm all for "making things work", I'm no longer willing to accomodate browsers that even Microsoft refuses to support anymore (see https://github.com/Pomax/react-onclickoutside/pull/12)

ken210:
Fair enough. For those that having same issue, I suggest using `require('classlist-polyfill');` once on top of your main file, before `require('react-onclickoutside')`. This way they will not have to fork this lib.

Pomax:
that's an excellent recommendation. I'll add that to the README.me

valentin-radulescu-hs:
I'd argue that IE9 is supported and that having React run perfectly in that browser but not this mixin is a bit of a pain. Just my 2c

Pomax:
perfectly valid. I'd argue that if you're on IE9, you're on an OS that has no reason not to be on the far more secure, far better IE11. Corporate restrictions might apply, but on a personal machine, upgrade IE first. You're *not* helping yourself by sticking with IE9, given how many things it lacks. IE10 was the first HTML5 browser, so at the very least, be on that. (I'll happily keep catching flack for only supporting HTML5 browsers)

valentin-radulescu-hs:
Wasn't looking for a lecture, I'm sure no one chooses freely to support IE :stuck_out_tongue:

Pomax:
no lecture intended, just giving you my 2ct back, since you indicated being interested in continuing the discussion by starting with "I'd argue" =)

PR #17, opened 2015-06-17
Fixed an issue with SVGElements in Safari

nikovanmeurs:
This commit solves an issue where the click handler throws an error because SVGElements do not contain a classList property in the Safari browser.

### DIFF
--- a/index.js
+++ b/index.js
@@ -50,7 +50,7 @@
           // thinking in terms of Dom node nesting, running counter
           // to React's "you shouldn't care about the DOM" philosophy.
           while(source.parentNode) {
-            found = (source === localNode || source.classList.contains(IGNORE_CLASS));
+            found = (source === localNode || source.classList && source.classList.contains(IGNORE_CLASS));
             if(found) return;
             source = source.parentNode;
           }


index.js
```diff
@@ -50,7 +50,7 @@
           // thinking in terms of Dom node nesting, running counter
           // to React's "you shouldn't care about the DOM" philosophy.
           while(source.parentNode) {
-            found = (source === localNode || source.classList.contains(IGNORE_CLASS));
+            found = (source === localNode || source.classList && source.classList.contains(IGNORE_CLASS));
```
Pomax: this moves the bug from "crashes on safari" to "does the wrong thing on safari". I'd rather we try to really fix this rather than merely making it not error out. If Safari supports classList but just not for the SVG element, we need to find out for which versions that is the case, and whether that happens when the page doctype indicates HTML5 content through `<!doctype html>` - as a true HTML5 element, if SVG is lacking .classList support then that is a bug in Safari that also needs to be filed on the Safari issue tracker, which I believe is straight up webkit.

jeffkole:
It appears that in Safari, the `classList` is defined on the `correspondingElement` that you can get from an [SVGElementInstance](http://www.w3.org/TR/SVG/struct.html#__svg__SVGElementInstance__correspondingElement).

If the code were changed to check for `correspondingElement && correspondingElement.classList`, would that be a reasonable solution?

Pomax:
if it were a secondary path check to see if the non-standard `correspondingElement` exist for an early return, followed by the real `classList` code, that would work.

jeffkole:
That jsbin does work on Safari for me, too (version 7.1.7 on desktop), but my use of SVG (and I'm guessing that @nikovanmeurs' use, too) is more complicated.  We use `<use/>` to reference other svg elements.  This jsbin more correctly demonstrates what is going on: http://jsbin.com/walujetimo/edit?html,js,output

When you click on the black parts of the alert icon, you will see output `[object SVGElementInstance] has classList? no`.  But if you click anywhere else, you will see a proper DOM hierarchy of elements that do have `classList`.  That first case is the one I am running into that is causing the errors.

jeffkole:
I received the same error (`Cannot call method 'contains' of undefined`) from our website when accessed on Android.  I do not have access to an Android device to reproduce it, but I suspect that the error was caused by the same underlying issue with SVG `use` elements.

Pomax:
Aha, so that's not the SVG element's classList not working, but `<use>` missing classList, which makes sense, since it does not technically reside in the rendered DOM. If someone wants to update this PR to check for the `correspondingElement` properly, with a comment pointing to needing to do this due to the SVG `use` element, probably with a link to http://www.w3.org/TR/SVG11/struct.html#InterfaceSVGUseElement in the comment too, then I'd be fine reviewing and landing that in.

PR #19, opened 2015-07-22
using the updated way to get the dom node on react

willgm:
The 'this.getDOMNode()' is deprecated and has been replaced with 'React.findDOMNode()'.

This point is also a impeditive to use it with the ES6 component that already don't has the getDOMNode.

I did a CodePen in onder to test it at least manually:
http://codepen.io/willgm/pen/oXPaov

### DIFF
--- a/index.js
+++ b/index.js
@@ -56,7 +56,7 @@
           }
           eventHandler(evt);
         }
-      }(this.getDOMNode(), this.handleClickOutside));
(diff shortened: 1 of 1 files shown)


Pomax:
thank you for the PR!

samternent:
This PR is a breaking change, unless React is defined globally, which cannot be assumed.

Issue: https://github.com/Pomax/react-onclickoutside/issues/20

PR #22, opened 2015-07-30
OnClickOutside for use with ES6 classes

marnusw:
I'm using [ES6 class React Components](https://facebook.github.io/react/blog/2015/01/27/react-v0.13.0-beta-1.html) which means mixins aren't supported. I'd like to propose the following to enable such use. It uses the same implementation as the mixin except that the component is passed into the various functions rather than them being member functions of the component itself. This also lends itself toward some configuration when initially registering the component.

I'd like to add some documentation to the main readme etc., but I'm expecting some discussion around this and would like to confirm your interest in this solution before going there. 

Excellent work on the mixin by the way!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ npm install react-onclickoutside --save
 
 (or `--save-dev` depending on your needs). You then use it in your components as:
 
-```
+```javascript
 var Component = React.createClass({
   mixins: [
     require('react-onclickoutside')
@@ -39,13 +39,13 @@ bower install react-onclickoutside
 
 and then include it as script via:
 
-```
+```html
 <script src="bower_components/react-onclickoutside/index.js"></script>
 ```
 
 Then use it as:
 
-```
+```javascript
 var Component = React.createClass({
   mixins: [
     OnClickOutside
@@ -66,7 +66,7 @@ When using this mixin, a component has two functions that can be used to explici
  
 In addition, you can create a component that uses this mixin such that it has the code set up and ready to go, but not listening for outside click events until you explicitly issue its `enableOnClickOutside()`, by passing in a properly called `disableOnClickOutside`:
 
-```
+```javascript
 var Component = React.createClass({
   mixins: [ ... ],
   handleClickOutside: function(evt) {
@@ -84,5 +84,42 @@ var Container = React.createClass({
 ## Marking elements as "skip over this one" during the event loop
 
 If you want the mixin to ignore certain elements, then add the class `ignore-react-onclickoutside` to that element and the callback won't be invoked when the click happens inside elements with that class.
+
+## ES6/2015 class support via HOC / ES7 decorators
+
+Since mixins can't be used with ES6/2015 class React components a 
+[Higher-Order Component (HOC)](https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750) 
+and [ES7 decorator](https://github.com/wycats/javascript-decorators) are bundled with the mixin: 
+
+```javascript
+import listensToClickOutside from 'react-onclickoutside/decorator'; 
+
+class Component extends React.Component {
+  handleClickOutside = (event) => {
+    // ...
+  }
+}
+
+export default listensToClickOutside(Component);
+
+// OR
+
+import listensToClickOutside from 'react-onclickoutside/decorator'; 
+
+@listensToClickOutside()
+class Component extends React.Component {
+  handleClickOutside = (event) => {
+    // ...
+  }
+}
+
+export default Component;
+```
+
+One difference when using the HOC/decorator compared to the mixin is that the `enableOnClickOutside()`
+and `disableOnClickOutside()` methods are not available as class methods, but rather on the `props`;
+so instead of `this.enableOnClickOutside()` you would call `this.props.enableOnClickOutside()`.
+
+In every other respect the the mixin and HOC/decorator provides the same functionality. 
 
 For bugs and enhancements hit up https://github.com/Pomax/react-onclickoutside/issues
--- a/decorator.js
+++ b/decorator.js
@@ -1,0 +1,97 @@
+var React = require('react');
+var objectAssign = require('react/lib/Object.assign');
+var OnClickOutsideMixin = require('react-onclickoutside');
+
+
+function addClickOutsideListener(Component) {
+
+  return React.createClass({
+
+    displayName: (Component.displayName || Component.name) + 'ClickOutside',
+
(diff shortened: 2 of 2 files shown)


Pomax:
I don't think it's wise to solve the problem around ES6 classes and mixins in this repo, while React itself hasn't solved this problem yet. Instead, I'd like the put up for consideration that it is more important that you file this solution on the React repositoriy on http://github.com/facebook/react/issues, so that they can have a look at what you're doing and whether that's a solution that they want to adopt as official "and here is how you do all this using ES6/ES2015". If there is consensus that this is a way forward for React, then I will be more than happy to accept PRs that implement the official solution.

I'll keep this PR open for now, because if they accept your solution then we can pretty much immediately land it, but I won't land it quite yet.

alexkuz:
@marnusw Why not just make a higher-order component for that, that provides `onClickOutside()` and other functions in props?

marnusw:
Thanks @alexkuz, based on your suggestion I have updated the solution with a higher-order component/wrapper function. This is **much** simpler, and I believe closer to the current *de facto* standard way of extending ES6 class components. I'll take it to the React team for comment from here if there is no further input.

Pomax:
go for it.

marnusw:
As I understand it the idea is that you build up the functionality of a specific component with wrappers within a single file, and then still use a single component in the parent; i.e. the wrapping is transparent outside the inner component's definition, consider:

```javascript
// FocussedSection.jsx

class FocussedSection extends React.Component {
  handleClickOutside = (event) => {
    // Remove Focus
  }
}

export default listenToClickOutside(FocussedSection);
```

```javascript
// SectionList.jsx

import FocussedSection from './FocussedSection.jsx'

class SectionList extends React.Component {
  render() {
    return (
      <div>
        {sections.map(section => <FocussedSection {...section}/>)}
      </div>
    );
  }
}
```
The outcome is exactly the same as applying the mixin to the `FocussedSection` component as far as I can see, and your example above would become:
```html
<ThingWrap>
  <ExtensionWrapping>
    <AuthFunctionality>
      // No OutsideClicks or VRSupport components here if they are both wrappers
      // inside the SecondaryUI file.
      <SecondaryUI>Not so bad</SecondaryUI>
    </AuthFunctionality>
  </ExtensionWrapping>
</ThingWrap>
```
Unless I'm missing something which is very possible...

The idea stems from [this blog post](https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750) originally and is also used by:
 * Yahoo in [Fluxbile](https://github.com/yahoo/fluxible): [See `connectToStores()` wrapper](http://fluxible.io/addons/connectToStores.html#example), which also supports the ES7 decorator pattern.
 * Material UI in their handling of [themes](http://material-ui.com/#/customization/themes) (see section 2. about half way down), where they also use a decorator.
 * I haven't checked behind the scenes, but [Relay](https://facebook.github.io/relay/) uses a similar syntax. Look at the last code block [in this example](https://facebook.github.io/relay/docs/tutorial.html#writing-the-game).

I think an ES7 decorator might well be the "neat thing to take advantage of" here. I could add decorator support to this PR as well!

inventorsspace:
@Pomax, given the most recent comments on facebook/react#5010, what are your thoughts? 

If you'd be willing to merge in principle I'll get this PR ready for merge. I can hardly remember the status of it, but I'm sure it still needs some docs etc.

Pomax:
In principle it's fine, but it completely changes how this thing works, so it'll basically be a deprecation of everything currently in the repo, with new docs and new code, since it won't be a mixin anymore.

marnusw:
I was thinking the two options could run in parallel? That is why I was glad that I eventually found a solution which is a minimal wrapper around the mixin allowing use as a decorator with a similar interface, i.e. you still only have to declare the `handleClickOutside` method, and apply either the mixin or decorator depending on preference.

Let me see if I can update the docs etc. in the next few days so it reflects the two options, then I'll check back to see what you think. Unless you definitely want to deprecate the mixin from your side. I think mixins are still very much in [common use](https://twitter.com/soprano/status/610867662797807617) though.

Pomax:
looking forward to that

marnusw:
I've updated the documentation to indicate how the higher-order component/decorator might complement the existing mixin rather than replace it. I would love to hear your thoughts.

You might note that I added one extra feature in the decorator which is as yet undocumented: the ability to pass a handler down on the `onClickOutside` prop from the parent rather than placing the handler in the child. I've found this useful on many occasions where the logic, for example, for showing a menu is maintained in a parent, but clicks outside the menu component should be detected. I'd like to leave that in, documented or undocumented, depending on your opinion.

I looked over `react-onoutsideevent` briefly and it looks like the API is quite similar although the implementation might be different. (I didn't look at the code.) That being the case these changes I propose can also be ported to that library once it has stabilized, and I would be happy to do so if you'll let me know when the time is right.

As a side note, I like the camel casing of the `disableOnClickOutside={true}` prop in this version since it feels more React-like with them having converted all events/attributes to camel cased versions. The new `disableonoutsideevent={true}` would be less so.

Pomax:
The outsideevent project (which is basically this mixin, but made generic) does use `listenToOutsideEvent` and `disableOnOutsideEvent` though.

.gitignore
```diff
@@ -0,0 +1,2 @@
+# IDE files
```
Pomax: hm, I'm not sure a .gitignore is necessary for these files - it's easy enough to pick them out of a PR.

Pomax:
If you can take out the `gitignore` then this looks good enough to merge to me - it might be worth filing a similar PR against the new generic outsideevent project; it will probably replace this mixin as recommended way to deal with outside clicks, but then "or other event types" rather than not having the option which event(s) count as outside-listenable.

marnusw:
Thanks @Pomax, the `.gitignore` is gone, I had only added it for convenience. This is good to merge then.

I'll put filing a PR on the outsideevent project on my to-do list. I might do so next weekend.

Pomax:
@marnusw: thanks! One last q, would it be possible to squash this to one commit? (just to make reverts easier should they end up being necessary)

marnusw:
Sure, I can do that. As far as I know if you do a merge via GitHub or using `git merge --no-ff`, so you get a new *merge commit* in master, removing that merge commit will pluck out all the commits from the feature branch as well; so squashing features to cater for reverts isn't strictly necessary. But I'll squash this for you in bit!

marnusw:
Squashed, ready to merge...

Pomax:
awesome, thanks!

PR #23, opened 2015-08-05
Checks for classList existence before proceeding

jeffkole:
In Safari, SVGElements do not have `classList` on them, so this component errors.  This is just a simple fix to not error out.


jeffkole:
Sorry. I didn't see #17 before.

PR #24, opened 2015-08-05
Fixes bug with SVG use not having classList

jeffkole:
This PR should replace #17

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,6 +35,22 @@
 
   var IGNORE_CLASS = 'ignore-react-onclickoutside';
 
+  var isSourceFound = function(source, localNode) {
+    if (source === localNode) {
+      return true;
+    }
+    // SVG <use/> elements do not technically reside in the rendered DOM, so
+    // they do not have classList directly, but they offer a link to their
+    // corresponding element, which can have classList. This extra check is for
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -35,6 +35,24 @@
...
+    // that case.
+    // See: http://www.w3.org/TR/SVG11/struct.html#InterfaceSVGUseElement
+    // Discussion: https://github.com/Pomax/react-onclickoutside/pull/17
+    if (source.correspondingElement) {
+      return source.correspondingElement.classList.contains(IGNORE_CLASS);
+    }
+    else {
+      return source.classList.contains(IGNORE_CLASS);
+    }
+  };
```
jeffkole: I pulled the logic into a function so that it would be more clear to read. If that isn't the style you like, I can put it inline instead.

Pomax: function is fine, but we don't need an `else` because of the early return:
```
if (source.correspondingElement) {
  return source.correspondingElement.classList.contains(IGNORE_CLASS);
}
return source.classList.contains(IGNORE_CLASS);
```

jeffkole:
That was amateur of me :) Fixed.

PR #25, opened 2015-08-08
Added ability to listen to the click event 

gmuresan:
I added the ability to listen to the click event instead of touchstart and mousedown by setting the useClickEvent property to true on your component.

There might be a cleaner way of implementing this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ When using this mixin, a component has two functions that can be used to explici
 
 - `enableOnClickOutside()` - Enables outside click listening by setting up the event listening bindings.
 - `disableOnClickOutside()` - Disables outside click listening by explicitly removing the event listening bindings.
- 
+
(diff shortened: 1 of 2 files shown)


Pomax:
got a link where I can read up on that? I don't think I've seen `useClickEvent` before

gmuresan:
@Pomax: I just mean you would add your own property called `useClickEvent ` in your component

Pomax:
hm. I'm hesitant to land this without a use-case to justify it: which problem did you run into that this patch solves?

gmuresan:
I have a navigation bar with a button that opens up a dropdown when you click on it. When the dropdown is open, I want to be able to click anywhere to close the dropdown. If I click down on the button that opened the dropdown closes, but when I release the mouse button, the click event gets sent to that button and the dropdown gets opened again.

Pomax:
I'm not sure I understand, that's what the `ignore-react-onclickoutside` class is for. Does your button use `<button className="... ignore-react-onclickoutside" .../>` so that when you click it, it doesn't double-trigger? If not: try that first.

Pomax:
closing due to inactivity

PR #26, opened 2015-08-20
Use ReactDom instead React to get rid of React warnings

valichek:


index.js
```diff
@@ -15,17 +15,17 @@
...
     // that support module.exports
-    module.exports = factory(require('react'));
+    module.exports = factory(require('react-dom'));
   } else {
     // Browser globals (root is window)
-    root.OnClickOutside = factory(React);
+    root.OnClickOutside = factory(ReactDom);
   }
-}(this, function (React) {
+}(this, function (ReactDom) {
```
Pomax: let's keep this `React` though

index.js
```diff
@@ -72,7 +72,7 @@
           }
           eventHandler(evt);
         }
-      }(React.findDOMNode(this), this.handleClickOutside));
+      }(ReactDom.findDOMNode(this), this.handleClickOutside));
```
Pomax: and the same here

index.js
eliseumds: The method *findDOMNode* is available at react-dom.

eliseumds:
Guys, this is a big update, isn't it? ReactDOM is released with React 0.14, so, besides increasing the major version, react >= 0.14 should be specified in *package.json*.

Pomax:
this hasn't landed. If it does, then this would be a major update, yes.

Pomax:
landed as major update, on npm as v4.0.1

PR #29, opened 2015-09-03
Adding check to prevent runtime error on IE

imtiyazj:
IE doesn't support classList property http://caniuse.com/#feat=classlist

### DIFF
--- a/index.js
+++ b/index.js
@@ -45,10 +45,16 @@
     // that case.
     // See: http://www.w3.org/TR/SVG11/struct.html#InterfaceSVGUseElement
     // Discussion: https://github.com/Pomax/react-onclickoutside/pull/17
-    if (source.correspondingElement) {
+    if (source.correspondingElement && source.correspondingElement.classList) {
       return source.correspondingElement.classList.contains(IGNORE_CLASS);
     }
-    return source.classList.contains(IGNORE_CLASS);
+
+    if (source.classList) {
+        return source.classList.contains(IGNORE_CLASS);
+    }
+    else {
+        return false;
+    }
   };
 
   return {


Pomax:
IE10 and IE11 should work just fine. Before any merging happens, can you show the code you're using that reveals a problem here still?

imtiyazj:
We have a custom select box where onclickoutside mixin is being used. The issue (javascript error) is happening only on older IE browsers (especially IE 8 and IE 9 in my case).

 var onClickOutside = require('react-onclickoutside');
...
mixins: [FluxibleMixin, onClickOutside]

Pomax:
The only reason you'd still need to target IE8 and 9 is for Windows XP users, and Windows XP is no longer supported by Microsoft so I really don't feel it makes sense to cater to anything lower than IE10.

If you absolutely need this functionality, you can fork it, patch it, and then point to your fork in your package.json?

imtiyazj:
We (at yahoo) still have to support IE 8/9 as they are available on Windows 7 etc and considered as A-grade browsers. As per market share, it is nearly 18% for these browsers https://www.netmarketshare.com/browser-market-share.aspx?qprid=2&qpcustomd=0

Maintaining code base at my end would work but will need high maintenance cost to keep it in sync with original source.

Pomax:
Then you have three months to do something clever: Microsoft is terminating any and all support for IE on Windows except "the latest version available for still supported platforms" (see [here](https://support.microsoft.com/en-us/lifecycle/search?sort=PN&alpha=internet%20explorer)). So, the the maintenance is going to be 90ish days; after that, anything older than IE11 is officially out of support.

eliseumds:
:+1: for not including IE9 code into the project. A shim should work just fine.

kennygrubb:
IE11 doesn't support classList on the <svg /> element and the polyfill suggested doesn't fix that either. This is a blocking issue for me using this module.

Pomax:
please file that as a new issue, this was about classList in general, IE11 implementing it for everything except SVG is a different thing to look at.

kennygrubb:
Thank you for the response.

It is actually a problem with the polyfill you guys recommend on the github
page. It doesn't account for svg elements in IE11.

Instead I used this polyfill which is based on the one you recommended but
adds support for svg in IE11... https://github.com/eligrey/classList.js/
... unfortunately, this project isn't published on npm, but it did fix the
issue.

Thanks,
Kenny

Pomax:
README.md updated to use that link instead.

kennygrubb:
That's great. I submitted a issue on the other github repo to take in that
fix so that the npm package will be updated.

https://github.com/yola/classlist-polyfill/issues/8

Kenny

PR #30, opened 2015-09-04
fix event bubbling up

chentsulin:
We will manually traverse up in the dom tree, so stop event bubbling up with `Event.stopPropagation()`


Pomax:
fair point

PR #31, opened 2015-09-11
Explicit check if handleClickOutside is a function

sverrejoh:
The error says function, so I think the condition should do it as well.


Pomax:
fair enough!

sverrejoh:
Thank you! An almost negligible improvement, but it looks nice :).

PR #33, opened 2015-09-18
Allow `disableOnClickOutside` as a state variable

sverrejoh:
Allowing `disableOnClickOutside` in state, so the parent component don't need to know about the inner mechanics of the component.

This can be used together with enableOnClickOutside/disableOnClickOutside if the whole component isn't part of the callout/dialog. (Which probably isn't good style in general, but sometimes you do what you got to do.

### DIFF
--- a/index.js
+++ b/index.js
@@ -81,7 +81,7 @@
 
       // If there is a truthy disableOnClickOutside property for this
       // component, don't immediately start listening for outside events.
-      if (!this.props.disableOnClickOutside) {
+      if (!this.props.disableOnClickOutside || !this.state.disableOnClickOutside) {
         this.enableOnClickOutside();
       }
     },


Pomax:
in that case we'll need to change this so that the conditional check only involves `this.state`, and bootstrap that using the construction properties:
```
getInitialState: function() {
  return {
    disableOnClickOutside: !!this.props.disableOnClickOutside
  };
},
...
```
However, that said whether or not the component is to be bypassed is the kind of thing that you, as UI designer, should be explicit about in the owning component. I'm not sure that letting the component decide this on its own is going to yield an as-intentional, as-obvious UI implementation.

Can you explain the use case you had in mind?

Pomax:
closing due to inactivity

PR #34, opened 2015-09-22
Adds support for browsers which don't support element.classList, by not using it at all

richardasaurus:
This makes it work in IE 10+11 (http://caniuse.com/#feat=classlist)


Pomax:
it already works in IE10 + 11. Unless you have a use case that demonstrates it doesn't, I have no intention of falling back to pre-HTML5 APIs just to please browsers that as of January 12th won't be supported by Microsoft anymore.

PR #36, opened 2015-10-07
use stopImmediatePropagation instead of stopPropagation

### LINKED ISSUE #35: Question: stop propagation when clicking outside?
Is there any way to to stop the event from bubbling when clicking outside? I tried something like
```
handleClickOutside: function(event) {
    event.stopPropagation()
}
```

I also tried the nativeEvent and other variations of this. It still seems to be triggering other onClick events, for example.

stevewillard:
Proposed change to fix #35

### DIFF
--- a/index.js
+++ b/index.js
@@ -58,7 +58,11 @@
 
       var fn = this.__outsideClickHandler = (function(localNode, eventHandler) {
         return function(evt) {
-          evt.stopImmediatePropagation();
+          if (evt.stopImmediatePropagation) {
+            evt.stopImmediatePropagation();
+          } else {
+            evt.stopPropagation();
+          }
           var source = evt.target;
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -58,7 +58,7 @@
 
       var fn = this.__outsideClickHandler = (function(localNode, eventHandler) {
         return function(evt) {
-          evt.stopPropagation();
+          evt.stopImmediatePropagation();
```
Pomax: better to do this as check + fn call, preserving stopPropagation, too:
```
if (evt.stopImmediatePropagation) {
  evt.stopImmediatePropagation();
} else {
  evt.stopPropagation();
}
```
because browser...

stevewillard:
Good idea! Updated.

Issue #40, opened 2015-10-11
Not worked properly when have multiple component using this mixin


kpaxqin:
I've leaved a comment under this issue:
https://github.com/Pomax/react-onclickoutside/issues/35

>I don't think it's a reasonable change, the stopImmediatePropagation will block all remaining listeners ! 
Just think we got two components(like datepicker) with react-onclickoutside on our page , the second one will never work properly because of the stopImmediatePropagation call in the first one's listener. 
And this is the problem I meet now: only the first datepicker can close the popup panel properly.
If you really want stopImmediatePropagation in some situations, maybe call it in the handleClickOutside is a better choice.

And I write a demo to show this issue, please note the second component never worked:
http://codepen.io/anon/pen/dYzPmM

and this is the same demo with react-onclickoutside@0.3.1 :
http://codepen.io/anon/pen/jbLExp

jbbr:
I can confirm this issue.
I'm using this mixin for a dropdown component to set the open state on a click outside event.

```
handleClickOutside = (event) => {
  console.log('Got click outside')
  this.setState({open: false});
};
```

With 0.3.2 state is set correctly. With 0.3.3 I get the `Got click outside` log but the state change has no effect at all.

Pomax:
reverted. v0.3.4 published.

PR #41, opened 2015-10-11
fix issue#40

kpaxqin:
see: https://github.com/Pomax/react-onclickoutside/issues/40 .


SpaceMacGyver:
This is confirmed to work for me in a situation where I have 4 separate components, in which only one of the four properly responds correctly.

PR #45, opened 2015-10-22
Support changing event type to listen to

austindebruyn:
I have a use case where I'd like to listen to outside mouse releases, not presses. There's many ways to do it, but the least intrusive is to just allow a prop that sets the evt type to listen on. Thoughts?
Tested and works fine on desktop and mobile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,21 +57,6 @@ var Component = React.createClass({
 });
 ```
 
-## Listen for mouse or touch release events
-
-By default, this mixin listens to mouse and touch press events. To instead listen for mouse and touch releases, set one or both of the props on your component called `outsideMouseEventType` and `outsideTouchEventType`.
-
-```
-  ...
-  getDefaultProps: function () {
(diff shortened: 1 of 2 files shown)


Pomax:
while a small change, that does change the mixin into a generic "onAnythingOutside", so might actually be worth just spinning off as its own, more generic, mixin.

austindebruyn:
I could live with it, but I'd rather not make a new repo that is 99% similar to this one. Are you a fan of this solution?

Pomax:
no, I don't want to have a project that's called "onclickoutside" lie about what it does. However, what we can do is start [react-onoutsideevent](https://github.com/Pomax/react-onoutsideevent), which I just have. If you want to refile your PR against that then we can continue work there, and eventually retire this project instead.

austindebruyn:
Fair enough, moved to https://github.com/Pomax/react-onoutsideevent/pull/1 .

PR #48, opened 2015-11-09
Fix React shallow renderer

pierlo-upitup:
This fix ensures that the document object is available before calling
methods on it. 
This allows using the React shallow renderer utilities with components that make use of this mixin.


Pomax:
thanks!

pierlo-upitup:
Any chance this could make its way into a `0.3.5` release? In our project we are currently still on React 0.13 and can't use ReactDOM. Thanks!

PR #55, opened 2015-11-19
remove root from global document

jslauthor:
Hi there. I couldn't get the decorator work in React 0.14.2 and tracked it down to root.document always evaluating to null. Not sure why "root" was included there as document is a global variable. This fixes it.


Pomax:
fair enough. merging and pushing 4.5

PR #56, opened 2015-11-20
Update index.js

deadivan:
fix "Uncaught ReferenceError: root is not defined" in my web app.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,6 +13,7 @@
  *
  */
 (function (root, factory) {
+  var root = this;
   if (typeof define === 'function' && define.amd) {
     // AMD. Register as an anonymous module.
     define(['react-dom'], factory);


Pomax:
The function already gets invoked with `this`, though: https://github.com/deadivan/react-onclickoutside/blob/deadivan-patch-1/index.js#L29

deadivan:
but root(Window object) did not defined here https://github.com/deadivan/react-onclickoutside/blob/deadivan-patch-1/index.js#L109
In my case , I use systemjs (in jspm) to manage my deps , does this related or caused to this problem?

deadivan:
Sorry ,my bad, i put it in wrong place. https://github.com/deadivan/react-onclickoutside/blob/deadivan-patch-1/index.js#L31
It should be here for https://github.com/deadivan/react-onclickoutside/blob/deadivan-patch-1/index.js#L110 to use

Pomax:
but `document` should be a global, since this is a "for the browser" library. It doesn't ever run without a document available - did you forget to tell your bundler to treat `document` as a global var?

deadivan:
yes,document is global here while root is not. Then root.document will cause "Uncaught ReferenceError: root is not defined" error.
```javascript
if (root.document != null) {
        document.removeEventListener("mousedown", fn);
        document.removeEventListener("touchstart", fn);
      }
```
line 1 use `root.document` , line 2 and line 3 just use `document`,not with `root`, after i defined `root` with `var root = this`, line 2 and line 3 won't cause error, so in my running environment,`document` is global .

Pomax:
but the current version does not have "root" there (https://github.com/Pomax/react-onclickoutside/blob/master/index.js) does this still happen for you on the latest?

PR #58, opened 2015-12-02
readme update for accurate browser support

vinnymac:
The ReadMe had me believing that this library would support IE9. But IE9 [does not support classList](http://caniuse.com/#search=classList). I checked IE9 and I had been getting `Unable to get value of the property 'contains': object is null or undefined`. Thanks for recommending the polyfill I will check it out.


Pomax:
thanks for verifying that. mergin.

PR #61, opened 2016-01-15
HOC option to expose selected public methods of the wrapped component.

marnusw:
The HOC wraps the original component so any public methods of that component are no longer be
available to its parent. This becomes problematic, especially, for example, when wrapping inputs
of which the `focus()` and `blur()` methods are often used. 

This adds the option to expose these methods on the HOC by supply the method names as an array on the second argument of the wrapper function or first argument of the decorator:

```javascript
  Child = listensToClickOutside(Child, ['focus', 'blur']);
  
  // OR
  
  @listensToClickOutside(['focus', 'blur'])
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,21 @@ class Component extends React.Component {
 export default Component;
 ```
 
+The HOC wraps the original component so any public methods of that component will no longer be
+available to its parent. This may be problematic, especially, for example, when wrapping inputs
+of which the `focus()` and `blur()` methods are often used. 
+
+To expose these methods on the HOC supply the method names as an array on the second argument of the 
+wrapper function or first argument of the decorator:
+
+```javascript
+  Child = listensToClickOutside(Child, ['focus', 'blur']);
+  
+  // OR
+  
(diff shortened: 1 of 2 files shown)


Pomax:
I've been thinking about this problem for other projects this week, as well, and as much as this solves the problem for focus and blur events, it doesn't solve the problem of the "decorator" effectively wiping away the true component's API. As a stopgap this PR works, but I'd much rather find an ES2015+ mixin replacement pattern that preserves access to the original component. 

I'd like to give this a little while to allow for research on what the best mixin replacement pattern is (in terms of API exposure as well as dev-friendliness). If we can't find one, this can land as "we couldn't solve this 'properly' within a reasonable timeframe", otherwise it hopefully becomes unnecessary due to a better pattern.

Pomax:
the code got a rewrite, so this PR won't merge in, but the new code is substantially less work to add this back into (no more mixin, just pure HOC). There is an issue open for this, but if you want to redo this PR on the updated v5 (major version because the release is backward compat. breaking) I'll be happy to merge it in.

PR #68, opened 2016-03-11
handle case when target DOM is detached

### LINKED ISSUE #63: handleClickOutside called because child element no longer in DOM
I have an issue where elements inside the `OnClickOutside` component are removed before the `isSourceFound` function is called.

Specifically, I'm using https://github.com/JedWatson/react-select (a dropdown component).

```javascript
<MyOnClickOutsideContainer>
  <ReactSelect />
</MyOnClickOutsideContainer>
```
Once a value is selected in the `<ReactSelect/>` component, then `handleClickOutside` is invoked. This is because `isSourceFound` looks to see if `event.target` is contained within `MyOnClickOutsideContainer`, however `event.target` **isn't** contained because it's already been removed from the DOM.

For the time being, I've done:
```javascript
  componentWillMount(){
    // this was happening because the react-select component was removing its contents before this event would fire.
    // by the time the event fired, the element was already removed from the DOM.
    this._oldHandleClickOutside = this.handleClickOutside;
    this.handleClickOutside     = (evt)=> {
      //http://stackoverflow.com/a/16820058/173957
      if (document.body.contains(evt.target)) {
        return this._oldHandleClickOutside(evt);
      }
    };
  },
```

chrisirhc:
Resolves #63

### DIFF
--- a/index.js
+++ b/index.js
@@ -73,6 +73,9 @@
             if(found) return;
             source = source.parentNode;
           }
+          // If element is in detached DOM, consider it not clicked
+          // outside as it can't be known whether it was outside.
+          if(source !== document) return;
           eventHandler(evt);
         }
       }(ReactDOM.findDOMNode(this), this.handleClickOutside));


chrisirhc:
Wow, that was quick. Thank you @Pomax :+1:

Issue #37, opened 2015-10-08
export IGNORE_CLASS from the module


goldensunliu:
instead of remembering the special className: 'ignore-react-onclickoutside' for ignoring click outside. It would be a better source truth if it was exported from the module and/or have an option to config your own className

Pomax:
can you show how you would see that used?

goldensunliu:
maybe something like this for an option to specify your own className
``` javascript
clickOutsideMixin = require('react-onclickoutside').mixin('ignore-clickoutside');
component = React.createClass({
  mixins: [require('react-onclickoutside').]
  ...
})
```
and now the className is 'ignore-clickoutside'

for the case of simply exporting IGNORE_CLASS:
``` javascript
ignoreClickoutsideClassName =  require('react-onclickoutside').IGNORE_CLASS
```

Pomax:
closing this issue, as the ignore class can be specified by devs who need that level of control.

PR #69, opened 2016-03-18
Allow specification of ignore class by prop

rafeememon:
Addresses #37

This allows consumers to get rid of the `'ignore-react-onclickoutside'` magic string in app code and use `IGNORE_CLASS` as the source of truth instead.

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,7 +37,7 @@
 
   var IGNORE_CLASS = 'ignore-react-onclickoutside';
 
-  var isSourceFound = function(source, localNode) {
+  var isSourceFound = function(source, localNode, ignoreClass) {
     if (source === localNode) {
       return true;
     }
@@ -48,9 +48,9 @@
     // See: http://www.w3.org/TR/SVG11/struct.html#InterfaceSVGUseElement
     // Discussion: https://github.com/Pomax/react-onclickoutside/pull/17
     if (source.correspondingElement) {
-      return source.correspondingElement.classList.contains(IGNORE_CLASS);
+      return source.correspondingElement.classList.contains(ignoreClass);
     }
-    return source.classList.contains(IGNORE_CLASS);
+    return source.classList.contains(ignoreClass);
   };
 
   return {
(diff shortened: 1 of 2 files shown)


Pomax:
Tather than exposing it without any further control, it would probably be more to have it be an `ignoreClass=...` property that can be passed down by the constructing parent. That way, if you -as developer- care about what the string reads, *you* control that, and if you don't then it doesn't matter what the code uses internally so keeing the fallback `ignore-react-onclickoutside` will be fine.

rafeememon:
Good suggestion -- in fact I've run into cases where I wanted different class names so that two elements listening to click-outside events could function independently.  I'll fix up this PR to incorporate that change.

rafeememon:
@Pomax, anything else I can do to help here?

Pomax:
man I really wish github told me about PRs being updated. I never got the notification that you posted a followup commit.

index.js
```diff
@@ -63,13 +63,14 @@
           evt.stopPropagation();
           var source = evt.target;
           var found = false;
+          var ignoreClass = this.props.clickOutsideIgnoreClass || IGNORE_CLASS;
```
Pomax: This probably needs to be mentioned in the README.md, too, before we land this, so people know they can use it. It might also be better to call this `outsideClickIgnoreClass` or `onClickOutsideIgnoreClass` (probably the first?)

Pomax:
@rafeememon small nits, I'll be happy to merge after those are addressed!

Pomax:
@rafeememon did you want to extend the readme with using this new property? If so, I can merge it in once that documentation has been added.

rafeememon:
Hey @Pomax, sorry for the radio silence, just got back from a mini-vacation.  I'll fix up the PR with your suggestions tomorrow.

Pomax:
gotcha. Hope it was a good one!

rafeememon:
Added a small blurb in the readme, and changed the prop name to `outsideClickIgnoreClass`.  Let me know if you think this needs anything else!

Pomax:
Nope, that's perfect, in it goes!

Pomax:
v4.7.0 published to npm

rafeememon:
Thanks for the merge!

PR #70, opened 2016-03-28
IE11 SVG's don't have classList or correspondingElement

mohicanlove:
SVG's in IE11 don't support classList or correspondingElement.
When clicking on an SVG it throws an error.
Test Case:
http://codepen.io/mohicanlove/pen/JXJzPE?editors=1111

This PR updates the `IGNORE_CLASS`  check so it returns false instead of raising an error when neither property is available.


Pomax:
If you need to rely on classlist, you don't need this specific library shimmed, you need *all* libraries shimmed, which you do by loading a shim that makes sure classList works for all HTML elements. You can find one on https://developer.mozilla.org/en/docs/Web/API/Element/classList

PR #73, opened 2016-03-30
Properly read outsideClickIgnoreClass from props

### LINKED ISSUE #72: Cannot read property 'outsideClickIgnoreClass' of undefined
Good library, but doesn't work 
```
Cannot read property 'outsideClickIgnoreClass' of undefined
```
Screenshot: http://joxi.ru/1A55PaaiKyqPyA
Code: 
```
import listensToClickOutside from 'react-onclickoutside/decorator';
@listensToClickOutside()
export default class Search extends Component {
....
  handleClickOutside = () => {
   alert('1');
  };
}
```

Can you help me?)

rafeememon:
Fixes #72

It turns out `this` is `document` in this context, not the component, which was causing the prop lookup to fail.  Not sure why this worked on my end before... sorry for the trouble @Pomax.

### DIFF
--- a/index.js
+++ b/index.js
@@ -58,12 +58,11 @@
       if(typeof this.handleClickOutside !== "function")
         throw new Error("Component lacks a handleClickOutside(event) function for processing outside click events.");
 
-      var fn = this.__outsideClickHandler = (function(localNode, eventHandler) {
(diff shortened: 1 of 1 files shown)


Pomax:
landed as v4.7.1

Issue #74, opened 2016-03-30
Need tests and CI


moodysalem:
The error in the most recent release could probably have been avoided

Pomax:
agreed, but not something I will have time for in the near future - if you know someone willing to add those, or want to take a stab at it yourself, I'll be more than happy to merge that work in.

rafeememon:
I'll take a stab at this.  Will report back

Pomax:
the test framework's been merged in, and while there is a low number of tests (two, right now), adding more tests based on bugs found will now be much easier (as in, possible, rather than not even possible)

PR #75, opened 2016-03-31
Add basic tests

rafeememon:
Addresses #74 

This sets up the test infrastructure with karma and mocha, and adds a few basic integration tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,25 @@
   "bugs": {
     "url": "https://github.com/Pomax/react-onclickoutside/issues"
   },
+  "scripts": {
+    "test": "karma start --single-run"
+  },
   "dependencies": {
     "object-assign": "^4.0.1"
+  },
+  "devDependencies": {
+    "chai": "^3.5.0",
+    "karma": "^0.13.22",
+    "karma-chai": "^0.1.0",
+    "karma-mocha": "^0.2.2",
+    "karma-phantomjs-launcher": "^1.0.0",
+    "karma-spec-reporter": "0.0.25",
+    "karma-webpack": "^1.7.0",
+    "mocha": "^2.4.5",
+    "phantomjs-prebuilt": "^2.1.7",
+    "react": "^0.14.8",
+    "react-addons-test-utils": "^0.14.8",
+    "react-dom": "^0.14.8",
+    "webpack": "^1.12.14"
   }
 }
--- a/test.js
+++ b/test.js
@@ -1,0 +1,31 @@
+var React = require('react');
+var ReactDOM = require('react-dom');
+var TestUtils = require('react-addons-test-utils');
+var OnClickOutside = require('.');
+
+describe('onclickoutside mixin', function() {
+
+  var OutsideClickComponent = React.createClass({
+    mixins: [OnClickOutside],
+
+    getInitialState: function() {
+      return {handleClickOutsideCalled: false};
+    },
+
+    handleClickOutside: function() {
+      this.setState({handleClickOutsideCalled: true});
+    },
+
+    render: function() {
+      return React.createElement('div');
+    }
+  });
+
+  it('should call handleClickOutside when clicking the document', function() {
+    var element = React.createElement(OutsideClickComponent);
+    var component = TestUtils.renderIntoDocument(element);
+    document.dispatchEvent(new Event('mousedown'));
+    assert(component.state.handleClickOutsideCalled);
+  });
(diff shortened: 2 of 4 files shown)


test.js
```diff
@@ -0,0 +1,38 @@
...
+    var element = React.createElement(OutsideClickComponent);
+    var component = TestUtils.renderIntoDocument(element);
+    ReactDOM.findDOMNode(component).dispatchEvent(new MouseEvent('mousedown'));
+    assert(!component.state.handleClickOutsideCalled);
+  });
+
+  it('should call handleClickOutside when clicking the document', function() {
+    var element = React.createElement(OutsideClickComponent);
+    var component = TestUtils.renderIntoDocument(element);
+    document.dispatchEvent(new MouseEvent('mousedown'));
```
rafeememon: I'm having a lot of issues dispatching events to things besides `document` and having them eventually bubble up.  Looking into it

Pomax: phantom.js?

rafeememon: I set this up to use Chrome (and hopefully Firefox/IE later) instead of phantomjs so that the tests could be more realistic.  Not sure if you have a preference for the direction things go here.

Pomax: mostly "not tied to any specific browser". I'm a big proponent of "it has to work on the internet" rather than "it has to work in browser X" so if this only works in Chrome, that's the the wrong way to go about it really... =/

The tests should pass irrespective of which browser's being used, as long as it conforms to modern APIs.

rafeememon: I'm sort or taking the mindset that works in Chrome, Firefox, and IE == works on the internet, and is slightly more realistic than phantomjs.  Haven't added the others yet because this is a WIP. =P  Haven't experimented with phantom here yet, but if that's also equivalent to "working on the internet" in your eyes then we should try that out.

Pomax: as long as it works in all three, that's good enough I'd say. I mostly suggested phantom because it seems pretty decent at "user testing" without a user (and automation is always a big win)

rafeememon: Right, but really the react test utils are giving us the automation; it's not unique to phantom.  I think we're in agreement here, but just wanted to be super sure, haha.

Pomax: sounds like we are =)

rafeememon:
Hey @Pomax, sorry for the lack of updates here.  I still haven't been able to properly simulate events that aren't dispatched directly to the document (event handler never gets called; obviously not a bug in onclickoutside code).  However I think there's still a lot of value in having just a basic test right now -- it would catch syntax errors and simple problems with the component.  Hopefully I or someone else can revisit adding more proper tests.  Let me know what you think.

Pomax:
basic tests and the prep-work that makes them run is better than nothing, so I'll review it today and probably land it in. I'll also be rewriting the library itself today as a major version update to clean some things up (most notably the HOC) so having the test code in will be a great help.

package.json
```diff
@@ -21,7 +21,25 @@
...
+    "karma-mocha": "^0.2.2",
+    "karma-phantomjs-launcher": "^1.0.0",
+    "karma-spec-reporter": "0.0.25",
+    "karma-webpack": "^1.7.0",
+    "mocha": "^2.4.5",
+    "phantomjs-prebuilt": "^2.1.7",
+    "react": "^0.14.8",
+    "react-addons-test-utils": "^0.14.8",
+    "react-dom": "^0.14.8",
+    "webpack": "^1.12.14"
```
Pomax: how is webpack used? (I don't see a webpack config file, so it's not clear how this is contributing to the testing as a whole)

rafeememon: Webpack packages up the test.js file (resolving requires etc) so that it can be run in phantom.  The karma-webpack plugin handles the config; all the normal webpack options are available but we don't need any of them here.

Pomax: ah, interesting. A little magical (not a fan of magic in code) but if it's a known "this is how karma/phantom work" then that'll do.

rafeememon: Agreed about magic; fortunately this is a pretty standard setup nowadays

rafeememon:
@Pomax, now that this is merged it may be a good idea to get CI set up, maybe [Travis](https://travis-ci.org/)?

Pomax:
travis is a pretty decent solution, I can see if I can add a travis.yml sometime today (unless you beat me to it of course)

Pomax:
https://github.com/Pomax/react-onclickoutside/issues/84

Issue #71, opened 2016-03-29
Document that calling disableOnClickOutside() in componentDidMount will not work


samsch:
As per the title, calling this.props.disableOnClickOutside() from componentDidMount() doesn't disable listening. This is using ES2015 class syntax (but not ES7/decorator).

Here is the relevent code snippet:
```
componentDidMount() {
    this.props.disableOnClickOutside();
}
```
With this code in my component, handleClickOutside is still being called.

[Here is a test case file.](https://gist.github.com/samsch/018bf78fd64d92396f5bb5f6c8583677)
It needs to be compiled with Babel+Webpack. If you want my exact config, you could drop the gist code into my [basic-react repo](https://github.com/samsch/basic-react) as `src/main.js` (the html grabs main.js, not main.jsx, but I used .jsx so that gist would highlight correctly). You would just need to do `npm install react-onclickoutside --save`, and `npm run dev`, then serve `public/` however.

~~I'm looking for a workaround as well. If I find one, I'll report back.~~
You can work around this by putting disableOnClickOutside in a setTimeout.
```
componentDidMount() {
    setTimeout(()=>this.props.disableOnClickOutside(), 0);
}
```

Pomax:
Creating an enabled component, but then immediately disabling it in the `componentDidMount` life cycle functions means you really should have built the component with the disable flag turned on instead. Using the code shown in the the [README.md](https://github.com/Pomax/react-onclickoutside#regulate-whether-or-not-to-listen-for-outside-clicks):

```
class App extends React.Component {
  ...
  render() {
    return ...
      <ABox clicks={this.state.numClicks} handler={this.handleClick} disableOnClickOutside={true} />
     ...;
  }
  ...
```

samsch:
That makes sense practically, but in a real life situation, I don't want to have to set an arbitrary property from my parent. This parent doesn't care if the child is listening or not.

I suppose this would lead to a solution of wrapping the HOCed component in another HOC which sets that property. This seems hacky. Am I missing another obvious solution here?

Pomax:
it's not an arbitrary property, though; the React philosophy is that parent control the bootstrapping of their children by specifying properties, and children control their state (potentially further bootstrapped on the `props` they received).

[...]

PR #76, opened 2016-04-06
Update docs to address disable listener call in componentDidMount difference for HOCs

samsch:
This adds documentation for the behaviour described in issue #71.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ var Container = React.createClass({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


Pomax:
I'd much rather be explicit about sensible programming here: if you need a component to mount in disabled state, literally do that by using `<... disableOnClickOutside={true}>`. Toggling state in a componentWill/DidMount is doing state wrong (any value set that way should have been a props-derived state)

samsch:
Ok, I drafted a different approach. I'm not sure if "anti-pattern" is too strong or not strong enough. I also considered something like "Using [disableOnClickOutside() in componentDidMount] is not part of the public API", but I'm not sure exactly how accurate that is.

Pomax:
anti-pattern's spot on: changing a component's state on mount means you should have bootstrapped the component properly instead.

Issue #65, opened 2016-02-03
disableOnClickOutside ES6/2015 class support


klassicd:
When using ES6 classes setting the property `<Component disableOnClickOutside={true} />`

Will overwrite the ability to use `this.props.disableOnClickOutside();` within the component class.  A simple solution I'm using now is to rename the property. 

```if (!this.props.disableOnClickOutside) {
        this.enableOnClickOutside();
}``` 

to 

```if (!this.props.disabled) {
        this.enableOnClickOutside();
}```

Pomax:
closing due to the release of v5.0.0 - the functionality is now a pure HOC solution with `props` pass-through, and README.md documentation that clarifies how to use its new form.

PR #77, opened 2016-04-07
Changed HOC/decorator to favor enable/disableClickOutside over this.props, rather than visa-versa.

samsch:
Fix for #65
Prior to this change, passing `enableOnClickOutside` or `disableOnClickOutside` as props to the HOC(component) will override the functions of the same name passed by the HOC to the wrapped component, which is not consistent with how the mixin works.

With this change, the internal function properties will now supersede the properties from the parent, allowing the component to enable/disable listening. Technically, this is a breaking change, but the documentation would seem to suggest it should work like this (the same as the mixin), rather than how it does currently.

### DIFF
--- a/decorator.js
+++ b/decorator.js
@@ -21,11 +21,11 @@ function addClickOutsideListener(Component) {
     },
 
     render: function render() {
-      return React.createElement(Component, objectAssign({
+      return React.createElement(Component, objectAssign({}, this.props, {
         enableOnClickOutside: this.enableOnClickOutside,
         disableOnClickOutside: this.disableOnClickOutside,
         ref: 'inner'
-      }, this.props));
+      }));
     }
   });
 }


decorator.js
```diff
@@ -21,11 +21,11 @@ function addClickOutsideListener(Component) {
     },
 
     render: function render() {
-      return React.createElement(Component, objectAssign({
+      return React.createElement(Component, objectAssign({}, this.props, {
```
Pomax: I'm slightly curious why this was set up the way it was, can we do a straight-up passthrough?
```
return React.createElement(Component, objectAssign({ref:'inner'}, this.props));
```
Since the HOC itself should not be doing anything with the properties that got set other than to just pass those on.

samsch: The enableOnClickOutside and disableOnClickOutside methods need to be added by the HOC to the properties passed to the component, or the component won't have access to them. If you replaced lines 24-28 with your snippet, then you wouldn't be able to enable/disable listening from the wrapped component.

Unless I'm misunderstanding your question?

Pomax: I may also be asking it wrong - the fact that the assignment order matters indicates that we two things that do completely different things with the same name, so that feels like a thing that needs to be fixed by chosing new names and releasing a new major version that breaks backward compatibility. And probably rewriting the HOC to then be a straight-up wrapper (with a props passthrough) instead of doing what it does right now.

samsch: Ah, yes. I agree fixing the naming conflict (as a breaking change in a major version) is probably the best path. To expose the enable/disable functions to the wrapped component would still require merging those properties with the passed in props, but there wouldn't be a naming conflict, so order wouldn't matter unless the user passed in props which have the same names as those functions.

Can this pull be merged as a fix for now?

Pomax:
I'd rather keep it open for 2 more days, I have Sunday reserved as a day to work on my issues backlog and this one ranks high

Pomax:
I've rewritten the code so that the package is now a pure HOC (no mixin), and pushed this as v5.0.0 to indicate it's absolutely backward compat. breaking

PR #78, opened 2016-04-08
Remove react/lib/objectAssign dependency.

thebuilder:
It's removed in React 15 - Polyfills like babel adds this.


dorsha:
This fix is important and required to be able to use React 15.x.
fixes: https://github.com/Pomax/react-onclickoutside/issues/79

Pomax:
@thebuilder when you say "Polyfills like babel adds this." do you mean no require is necessary at all, or is a require still necessary, just for a generic not-react library?

rafeememon:
This can probably be closed since #80 was merged

Issue #79, opened 2016-04-08
React 15 and Object.assign


rmdort:
React 15 doesnt ship with Object.assign . Getting module not found error

````
Module not found: Error: Cannot resolve module 'react/lib/Object.assign' in ..../node_modules/react-onclickoutside
 @ ./~/react-onclickoutside/decorator.js 2:19-53
````

anorudes:
+1

mijamo:
Related to this: https://github.com/facebook/react/pull/6376

jshthornton:
I've included a fix in this PR: https://github.com/Pomax/react-onclickoutside/pull/80

Pomax:
landed, v4.8.0 pushed to npm

PR #80, opened 2016-04-08
Adding react v15 support

jshthornton:
#79


Pomax:
what does this do for people still on React 14 though? I'll happily merge in code that makes things work with the latest React, but we don't want to break people who can't switch to 15 for a while.

jshthornton:
No changes, as it is uses an existing well established library.

Pomax:
in that case in it goes, thanks for the fix!

Pomax:
v4.8.0 pushed to npm

PR #81, opened 2016-04-11
prevent undefined error if called server side

Sinewyk:
does the same thing as before (not attaching handlers)
except it doesn't throw anymore if there is no document


Pomax:
good call, not sure why those were `null` checks in the first place, they should have definitely been typeof undef checks! Thanks for the PR

Pomax:
v4.9.0 pushed to npm

PR #99, opened 2016-05-23
Update index.js to use click event and not mousedown

sparksm:
if the state of a component changes based on the outside click we want the target area to receive the completed event before modifying state.

take an accordion menu component for example - if an area above the next target collapses, we want to wait until the event is complete so that the next item can expand as a result of state change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -156,6 +156,7 @@
           var fn = this.__outsideClickHandler;
           if (typeof document !== "undefined") {
             var events = this.props.eventTypes || DEFAULT_EVENTS;
+            if (!events.forEach) { events = [events] };
             events.forEach(function (eventName) {
               document.addEventListener(eventName, fn);
             });
@@ -170,6 +171,7 @@
           var fn = this.__outsideClickHandler;
           if (typeof document !== "undefined") {
             var events = this.props.eventTypes || DEFAULT_EVENTS;
+            if (!events.forEach) { events = [events] };
             events.forEach(function (eventName) {
               document.removeEventListener(eventName, fn);
             });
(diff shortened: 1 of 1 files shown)


Pomax:
hm, rather than land this, it'd be better to make the event trigger a custom option then, like `eventName='...'` and then taking either string or array of strings

sparksm:
sounds good. leveraging optional `eventTypes` array to props and default to `["mousedown", "touchstart"]`

index.js
```diff
@@ -154,8 +154,10 @@
         enableOnClickOutside: function() {
           var fn = this.__outsideClickHandler;
           if (typeof document !== "undefined") {
-            document.addEventListener("mousedown", fn);
-            document.addEventListener("touchstart", fn);
+            var events = this.props.eventTypes || ["mousedown", "touchstart"];
```
rafeememon: It might be a good idea to pull the default event types into a "constant" at the top of the file

sparksm: agreed. the whole thing could be a bit DRYer.

Pomax:
nice! although `forEach` will break if the passed value is a string, which is a good idea to cover in the interest of "write only what you need" from a user point of view. If they only need "click", then making them write `eventTypes={['click']}` instead of `eventTypes='click'` is making them do more work than is strictly necessary, so can we add a small bootstrapping to the `events ` extraction?
```
var events = this.props.eventTypes || DEFAULT_EVENTS;
if (!events.forEach) events = [events];
events.forEach({
  ...
});
```

sparksm:
Gladly added.

Pomax:
awesome! I'll merge this in and push a new version, after updating the README.md to expose this new information.

Pomax:
v5.2.0 published to npm

PR #100, opened 2016-05-24
Fixed issue where React v15.1.0 would warn against binding a component method

mr-winter:


Pomax:
This fixes the warning, by changing the behaviour of the bind completely. `bind` takes, technically, an unlimited number of arguments, with the first argument being the reference to what the `this` context will be when the derived function is called, and any subsequent argument being a prebound call argument, so given `somefunc() { return this.y; }` and `var fn = somefunc.bind(x);`, the result of this binding is that `fn()` effectively return `x.y` because `this` points to `x`. If we change this to `somefunc.bind(null,x)`, we completely lose that, and instead `fn()` will return undefined, because `this` points to null context, which is global context.

PR #102, opened 2016-06-15
Avoid `console.warn` via delegation, fixes #89

### LINKED ISSUE #89: React 15.0.1 - warning about binding
Hi,
I've recently updated React in my app to version 15.0.1 and since then, this warning has been poping up in console: 

`app.js:5245 Warning: bind(): You are binding a component method to the component. React does this for you automatically in a high-performance way, so you can safely remove this call. See Dropdown`

When I remove react-onclickoutside implementation in my Dropdown component, the warning disappears, obviously. Are you planning to address this issue?

Thanks

craigdallimore:
Hello!

We were also experiencing the issues reported in #89. It looks like we can avoid that warning by providing the context to `handleClickOutside` in a way that `React` doesn't take issue with.

Tests output before and after:

screen shot 2016-06-15 at 17 14 31

Let me know if there are any changes required!

### DIFF
--- a/index.js
+++ b/index.js
@@ -60,6 +60,7 @@
     }
   };
 
+
   /**
    * This function generates the HOC function that you'll use
    * in order to impart onOutsideClick listening to an
--- a/index.js
+++ b/index.js
@@ -60,16 +60,6 @@
     }
   };
 
-  /*
-   * A workaround for issue #89, where React will emit a console.warn when it
-   * detects methods bound using `.bind`
-   */
-  var delegateHandler = function(instance) {
-    return function() {
-      instance.handleClickOutside.apply(instance, arguments);
-    }
-  };
-
   /**
    * This function generates the HOC function that you'll use
    * in order to impart onOutsideClick listening to an
@@ -115,7 +105,7 @@
 
           var fn = this.__outsideClickHandler = generateOutsideCheck(
             ReactDOM.findDOMNode(instance),
-            delegateHandler(instance),
+            instance.handleClickOutside.bind(instance, undefined),
             this.props.outsideClickIgnoreClass || IGNORE_CLASS,
             this.props.preventDefault || false,
             this.props.stopPropagation || false
--- a/index.js
+++ b/index.js
@@ -60,6 +60,15 @@
     }
   };
 
+  /*
+   * A workaround for issue #89, where React will emit a console.warn when it
+   * detects methods bound using `.bind`
+   */
+  var delegateHandler = function(instance) {
+    return function() {
+      instance.handleClickOutside.apply(instance, arguments);
+    }
+  };
 
   /**
    * This function generates the HOC function that you'll use
@@ -104,16 +113,9 @@
             throw new Error("Component lacks a handleClickOutside(event) function for processing outside click events.");
           }
 
-          // react@15.x will emit warnings when methods are bound using the
-          // .bind() syntax; this provides a way to delegate to
-          // `handleClickOutside` without React emitting a `console.warn`.
-          var delegateHandler = function() {
-            instance.handleClickOutside.apply(instance, arguments);
-          }
-
           var fn = this.__outsideClickHandler = generateOutsideCheck(
             ReactDOM.findDOMNode(instance),
-            delegateHandler,
+            delegateHandler(instance),
             this.props.outsideClickIgnoreClass || IGNORE_CLASS,
             this.props.preventDefault || false,
             this.props.stopPropagation || false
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -104,9 +104,16 @@
...
+          // .bind() syntax; this provides a way to delegate to
+          // `handleClickOutside` without React emitting a `console.warn`.
+          var delegateHandler = function() {
+            instance.handleClickOutside.apply(instance, arguments);
+          }
+
           var fn = this.__outsideClickHandler = generateOutsideCheck(
             ReactDOM.findDOMNode(instance),
-            instance.handleClickOutside.bind(instance),
+            delegateHandler,
```
Pomax: `bind` should only cause a problem inside the react class, so a better solution might be to have a top level
```
function delegateHandler(instance) {
  return instance.handleClickOutside.bind(instance);
}
```
and then here use:
```
... = ... = generateOutsideCheck(
  ReactDome.findDOMNode(instance),
  delegateHandler(instance),
  ...
);
```

craigdallimore:
I agree; a top level function is a nicer solution. I just tried what you have proposed, I'm afraid `react` still detected the binding created with `.bind`.

This worked out, may I offer this instead?

```
var delegateHandler = function(instance) {
  return function() {
    instance.handleClickOutside.apply(instance, arguments);
  };
};
```
...
```
var fn = this.__outsideClickHandler = generateOutsideCheck(
              ReactDOM.findDOMNode(instance),
             instance.handleClickOutside.bind(instance),
             delegateHandler(instance),
             ...
```

Pomax:
hm, that'll still generate a new function every call, so in that case keeping it the original code is fine... the only other thing I can think of is an import...
```
var delegateHandler = require('./delegateHandler');
```
with the `delegateHandler.js` file containing that agnostic `module.exports = function delegateHandler(instance) { return instance.handleClickOutside.bind(instance); }`; function.

craigdallimore:
Looking into this, I found this function in `/node_modules/react/lib/ReactClass.js`.

If I understand it correctly, it will call `.bind` on the `handleClickOutside`, passing a reference to the component as context, then replaces it's `.bind` with a custom version which will
- check it's `newThis` matches the component (it does)
- looks for the presence of arguments (we supply none in our `instance.handleClickOutside.bind(instance)`)

```
function bindAutoBindMethod(component, method) {
  var boundMethod = method.bind(component);
  if (process.env.NODE_ENV !== 'production') {
    boundMethod.__reactBoundContext = component;
    boundMethod.__reactBoundMethod = method;
    boundMethod.__reactBoundArguments = null;
    var componentName = component.constructor.displayName;
    var _bind = boundMethod.bind;
    boundMethod.bind = function (newThis) {
      for (var _len = arguments.length, args = Array(_len > 1 ? _len - 1 : 0), _key = 1; _key < _len; _key++) {
        args[_key - 1] = arguments[_key];
      }

      // User is trying to bind() an autobound method; we effectively will
      // ignore the value of "this" that the user is trying to use, so
      // let's warn.
      if (newThis !== component && newThis !== null) {
        process.env.NODE_ENV !== 'production' ? warning(false, 'bind(): React component methods may only be bound to the ' + 'component instance. See %s', componentName) : void 0;
      } else if (!args.length) {
        process.env.NODE_ENV !== 'production' ? warning(false, 'bind(): You are binding a component method to the component. ' + 'React does this for you automatically in a high-performance ' + 'way, so you can safely remove this call. See %s', componentName) : void 0;
        return boundMethod;
      }
      [elided]
}
```

As another approach, I offer this:

```
          var fn = this.__outsideClickHandler = generateOutsideCheck(
            ReactDOM.findDOMNode(instance),
            instance.handleClickOutside.bind(instance, undefined),
            this.props.outsideClickIgnoreClass || IGNORE_CLASS,
            this.props.preventDefault || false,
            this.props.stopPropagation || false
          );
```

index.js
```diff
@@ -106,7 +106,7 @@
 
           var fn = this.__outsideClickHandler = generateOutsideCheck(
             ReactDOM.findDOMNode(instance),
-            instance.handleClickOutside.bind(instance),
+            instance.handleClickOutside.bind(instance, undefined),
```
Pomax: That's an awesome fix.

Pomax:
sorry it took a while to merge in, I've been on a fairly intense workweek. Awesome fix though, spinning up a new release now.

Pomax:
v5.3.0 published to npm

craigdallimore:
Fantastic, thank you.

PR #107, opened 2016-06-27
Add enable/disableOnClickOutside prop. Fixes #106

### LINKED ISSUE #106: Missing disable/enable onClickOutside function props?
Are disableOnClickoutside() and enableOnClickOutside() still supposed to be passed as props to the wrapped component in 5.x? As far as I can tell, they are not being passed down.

samsch:
Fixes https://github.com/Pomax/react-onclickoutside/issues/106

This change adds enableOnClickOutside and disableOnClickOutside props back to the wrapped component, equivalent to the 4.x decorator/HoC version of the library.

This isn't the prettiest solution, in my own code I would use Object.assign. But, this code will work where Object.assign doesn't exist.

In the future, should this library be converted to use an ES2015 class component rather than createClass, this bit of code will need to be changed to bind the functions to `this` before adding them to the wrapped component's props.

### DIFF
--- a/index.js
+++ b/index.js
@@ -106,7 +106,7 @@
 
           var fn = this.__outsideClickHandler = generateOutsideCheck(
             ReactDOM.findDOMNode(instance),
-            instance.handleClickOutside.bind(instance, undefined),
+            instance.handleClickOutside.bind(instance),
             this.props.outsideClickIgnoreClass || IGNORE_CLASS,
             this.props.preventDefault || false,
             this.props.stopPropagation || false
--- a/index.js
+++ b/index.js
@@ -183,10 +183,13 @@
          */
         render: function() {
           var passedProps = this.props;
-          var props = { ref: 'instance' };
+          var props = {};
           Object.keys(this.props).forEach(function(key) {
             props[key] = passedProps[key];
           });
+          props.ref = 'instance';
+          props.disableOnClickOutside = this.disableOnClickOutside;
+          props.enableOnClickOutside = this.enableOnClickOutside;
           return React.createElement(Component,  props);
         }
       });
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-onclickoutside",
-  "version": "5.3.1",
+  "version": "5.3.2",
   "description": "An onClickOutside mixin for React components",
   "main": "index.js",
   "homepage": "https://github.com/Pomax/react-onclickoutside",
--- a/test/test.js
+++ b/test/test.js
@@ -12,7 +12,11 @@ describe('onclickoutside hoc', function() {
       };
     },
 
-    handleClickOutside: function() {
+    handleClickOutside: function(event) {
+      if (event === undefined) {
+          throw new Error("event cannot be undefined");
+      }
+
       this.setState({
         clickOutsideHandled: true
       });
--- a/test/test.js
+++ b/test/test.js
@@ -5,11 +5,29 @@ var wrapComponent = require('../index');
 
 describe('onclickoutside hoc', function() {
 
+  var externalToggleEnableClickOutside;
+
   var Component = React.createClass({
     getInitialState: function() {
       return {
-        clickOutsideHandled: false
+        clickOutsideHandled: false,
+        timesHandlerCalled: 0
       };
+    },
+
+    toggleEnableClickOutside: function(on) {
+      if(on) {
+        this.props.enableOnClickOutside();
+      }
+      else {
+        this.props.disableOnClickOutside();
+      }
+    },
+
+    componentDidMount: function() {
+      externalToggleEnableClickOutside = function(on) {
+        this.toggleEnableClickOutside(on);
+      }.bind(this);
     },
 
     handleClickOutside: function(event) {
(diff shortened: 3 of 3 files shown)


Pomax:
Hey, sorry for the late review - would it be possible to also extend the test file so that it breaks without your fix, but works with this fix in it?

samsch:
Ok, technically I did what you asked. This is my first shot at karma, so you'll want to take a look over it before merging it. Style-wise, I re-used the test component by adding a couple methods. I setup an internal trigger by using a variable scoped to the test running function. I'm guessing there is a better way to do that, but I'm just not that familiar/skilled with React component testing.

There is also a warning from React about binding a method. Unless I've got something else wrong, the bind that I do *is* required to run the tests, since the function created in componentDidMount would not otherwise reference the component instance as `this` when called externally.

Anyhow, this test fails with "this.props.disableOnClickOutside() should not be undefined." when run against 5.3.2, and succeeds when run against this pull.

test/test.js
```diff
@@ -52,4 +75,26 @@ describe('onclickoutside hoc', function() {
...
+
+  it('should not call handleClickOutside if this.props.disableOnClickOutside() is called, until this.props.enableOnClickOutside() is called.', function() {
+    var element = React.createElement(WrappedComponent);
+    var component = TestUtils.renderIntoDocument(element);
+    document.dispatchEvent(new Event('mousedown'));
+    var instance = component.getInstance();
+    assert(instance.state.timesHandlerCalled === 1, "handleClickOutside called");
+
+    try {
+      externalToggleEnableClickOutside(false);
```
chrisirhc: You should be able to simply call `instance.toggleEnableClickOutside()`

test/test.js
```diff
@@ -5,16 +5,39 @@ var wrapComponent = require('../index');
...
+
+    handleClickOutside: function(event) {
+      if (event === undefined) {
+          throw new Error("event cannot be undefined");
+      }
+
       this.setState({
-        clickOutsideHandled: true
+        clickOutsideHandled: true,
+        timesHandlerCalled: this.state.timesHandlerCalled + 1
```
chrisirhc: Instead of using `state`, you can use a simple variable outside similar to how you're using `externalToggleEnableClickOutside`.

chrisirhc:
Ran into this today, so I did a quick review.

test/test.js
samsch: I just used state for consistency with the original. Either way is fine with me, though if one var is pulled from state, they probably both should be.

test/test.js
samsch: I implemented this change. Shaves off the need for componentDidMount. I made the test seem cleaner.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-onclickoutside",
-  "version": "5.3.1",
+  "version": "5.3.2",
```
chrisirhc: Might want to rebase your changes on top of master so these don't show up here.

index.js
```diff
@@ -183,10 +183,13 @@
...
         render: function() {
           var passedProps = this.props;
-          var props = { ref: 'instance' };
+          var props = {};
           Object.keys(this.props).forEach(function(key) {
             props[key] = passedProps[key];
           });
+          props.ref = 'instance';
+          props.disableOnClickOutside = this.disableOnClickOutside;
+          props.enableOnClickOutside = this.enableOnClickOutside;
```
chrisirhc: Alternatively, set this up above, in `props = {}`. I think it's okay to have them overridden if the user chooses to do so.

samsch: Actually, we can't, at least for disableOnClickOutside. If you pass a prop of that name to the wrapped component, it's the default disable. I'm not a huge fan of them having the same name, but it would be a major backward compatibility break to change that now, and it doesn't have a significant benefit to do so that I can see.

samsch:
I've been pretty busy this week, and I expect next week too, so if you (or someone else) want(s) to take over the pull for more changes (rebasing), go ahead.

test/test.js
Pomax: As this is a testing-only value it's not super important,  but it's probably best to lift this out of the state - even if we're just testing, updating the state in any way causes React to run through the `render()` pipeline, which we probably want to avoid.

Pomax:
I've been dealing with compound hand injuries for the past 2 weeks, but am finally able to get back to my github queue - I'll see if I can pick up this PR if you're strapped for time.

klacointe:
:+1:

Pomax:
hi @samsch, sorry for the long pause - I just picked up some outstanding PRs again and this PR will need a rebase due to another PR landing that touched some lines that your commits also touch.

As for the version number update: best to leave that alone, the `npm version` command does semver updates in package.json and git tagging automatically, so no need to change that in PRs (it gets run after master is updated from a PR getting merged in)

radufilipescu:
:+1:

samsch:
@Pomax I've mostly fallen out of the loop for the new changes since I worked on this. If you (or someone else) want to take over the pull and update/replace/rebase it as needed, that would be great.

Pomax:
same here, I'll take some time today to rebase and and update where necessary.

Pomax:
superceded by #122

Pomax:
v5.4.0 released

PR #110, opened 2016-07-06
This reverts #102 with a test why it was broken

tcoopman:
This commits reverts #102 (so there are console.warns) again, but the PR was broken (see test). So I believe the revert is the sensible thing todo.


Pomax:
yeah this will have to do - code can live with the warning (it still works just fine, it's just a blanket React notice) but not with the event handler getting called incorrectly.

PR #112, opened 2016-07-11
fix react warning

tcoopman:


index.js
```diff
@@ -106,7 +106,8 @@
 
           var fn = this.__outsideClickHandler = generateOutsideCheck(
             ReactDOM.findDOMNode(instance),
-            instance.handleClickOutside.bind(instance),
+            instance.handleClickOutside,
```
Pomax: We might as well only pass in the instance, then, and the refer to `instance.handleClickOutside` rather than `eventHandler` inside the `generateOutsideCheck` code.

tcoopman: fixed

imdpk:
Any update on when this fix will be merged in?

Pomax:
I've had compound hand injuries to deal with for the last two weeks, so I'm only just now able to get back to visiting my github work queue.

Pomax:
tests pass, warning's disappeared, in it goes.

Pomax:
v5.3.3 published to npm

imdpk:
Beautiful thank you!!

PR #113, opened 2016-07-11
move react to a peerDependency, this makes it easier to use npm link

tcoopman:


Pomax:
Not quite sure I understand the use case, how does this make using `npm link` easier?

package.json
```diff
@@ -37,9 +37,11 @@
     "karma-webpack": "^1.7.0",
     "mocha": "^2.4.5",
     "phantomjs-prebuilt": "^2.1.7",
-    "react": "^15.0.0",
```
rafeememon: `peerDependencies` aren't automatically installed in npm v3+, so this probably needs to stay in `devDependencies` as well so they don't need to be manually installed during development

tcoopman: The problem with *npm link* and *react*: https://gist.github.com/jimfb/4faa6cbfb1ef476bd105#gistcomment-1686426

Also, if you use react-onclickoutside, then you will have react installed, no?

Pomax: That depends on what you're asking about - devDependencies are only for when you're working *on* the code, issuing `npm install` specifically in its own dir, not for when you're using the code as a dependency in a larger project. When npm installs something as a normal library installation, the devDependencies don't resolve at all, so you should be able to use `npm link` in your personal React clone's dir and as a normal library react-onclickoutside should now be using your React, not the one indicated in the package.json.

Only if you do dev work on this code, will it use the "regular" React because the package.json instructs it to explicitly install it as full project install.

So I'd still like to know the exact use case *you* ran into (what were you trying to do/how were you using the library) that necessitates moving the dependency to a peerDependency, which makes dev work considerably more annoying because it will force people who work on this code to always manually install `react` and `react-dom` in addition to running `npm install` (peer dependencies are *not* installed by npm anymore since v3)

tcoopman: The use case for this is developing this component with `npm link`. The problem is that in the other case you can have 2 versions of react and react doesn't like that.

To me it looks like it only complicates things for people who want to develop react-onclickoutside. Anyone who uses this, will already have react.

I will close this at the moment, because it looks like I'm the only one with the issue at the moment.

PR #114, opened 2016-07-17
Add support for config, allow users to specify onClickOutside handler

### LINKED ISSUE #115: Add support for config, allow users to specify onClickOutside handler
Add the support of configuration of the HOC. The most apparent setting is to allow the users to specify which method of their components should be called when the click outside happens. The reason for this is that it removes the implicit dependency on the `handleClickOutside` method. This means that if my component does not have this method, the error is manifested at runtime, whereas if I explicitly specify the handler, the error can be detected in compile-time (e.g. when using TypeScript). For more on implicit dependencies I recommend this [article on mixins and HOC's](https://facebook.github.io/react/blog/2016/07/13/mixins-considered-harmful.html).

The support for config would also allow to introduce more configuration options or moving some configuration from props to config (e.g. `eventTypes`) in the future.

no23reason:
Fixes #115.

The implementation maintains compatibility with the previous call convention (as demonstrated in the tests), so it should not break semver.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,27 @@ var MyComponent = onClickOutside(React.createClass({
 }));
 
 ```
+Note that if you try to wrap a React component class without `handleClickOutside(evt)` handler like this, the HOC will throw an error.
+If you want to execute method different than `handleClickOutside(evt)`, you can specify it as a second parameter to `onClickOutside` like this:
 
-Note that if you try to wrap a React component class without `handleClickOutside(evt)` handler, the HOC will throw an error. If you want onClickOutside functionality, you *must* have this function defined.
+```javascript
+// load the HOC:
+var onClickOutside = require('react-onclickoutside');
+
+// create a new component, wrapped by this onclickoutside HOC:
+var MyComponent = onClickOutside(React.createClass({
+  ...,
+  myClickOutsideHandler: function(evt) {
+    // ...handling code goes here...
+  },
+  ...
+}), function (instance) {
+    instance.myClickOutsideHandler
+});
+
+```
+
+Note that if you try to wrap a React component class while specifying a custom handler and the component does not implement it, the HOC will throw an error.
 
 ## Regulate which events to listen for
 
--- a/README.md
+++ b/README.md
@@ -19,44 +19,32 @@ $> npm install react-onclickoutside --save
 var onClickOutside = require('react-onclickoutside');
 
 // create a new component, wrapped by this onclickoutside HOC:
-var MyComponent = onClickOutside(
-    // ...config can go here...
-)(React.createClass({
+var MyComponent = onClickOutside(React.createClass({
   ...,
-  handleClickOutside: function (evt) {
+  handleClickOutside: function(evt) {
     // ...handling code goes here...
   },
   ...
 }));
 
-// to maintain compatibility with previous versions, the wrapper can be also called like this:
-var MyLegacyComponent = onClickOutside(React.createClass({
-  ...,
-  handleClickOutside: function (evt) {
-    // ...handling code goes here...
-  },
-  ...
-}));
 ```
 Note that if you try to wrap a React component class without `handleClickOutside(evt)` handler like this, the HOC will throw an error.
-If you want to execute method different than `handleClickOutside(evt)`, you can specify it using the config parameter of `onClickOutside` like this:
+If you want to execute method different than `handleClickOutside(evt)`, you can specify it as a second parameter to `onClickOutside` like this:
 
 ```javascript
 // load the HOC:
 var onClickOutside = require('react-onclickoutside');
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -19,17 +19,48 @@ $> npm install react-onclickoutside --save
 var onClickOutside = require('react-onclickoutside');
 
 // create a new component, wrapped by this onclickoutside HOC:
-var MyComponent = onClickOutside(React.createClass({
+var MyComponent = onClickOutside(
```
Pomax: this gets unwieldy. Making the function take an optional object would be a better API, making it support both `onClickOutside(ReactComponent)` calls as well as `onClickOutside(ReactComponent, optionsObject)`

Pomax:
Can this be given an issue, as well, outlining the currently problem that can be fixed by passing in a config object, and that this PR can then be updated with (putting a `fixes #....` in the PR comment, so github takes care of the cross-linking and auto-closing if the PR lands)

README.md
no23reason: I had it similarly to what you are proposing in a previous version, but I decided to go with the idiom used in other HOC's such as [connect in react-redux](https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options).

If you really want, I can change it.

Pomax: That looks a little different in terms of use-case, though - I'd rather we change this such that `onClickOutside` has the signature `onClickOutside(componentClass, [options])`, maintaining backward compatibility.

Pomax:
I was out of commission for two weeks due to compound hand injuries, just getting back to my issue list - a small fix was landed into master that fixes the whole bind warning nonsense (finally), so this PR will need a small rebase for that.

no23reason:
I reverted to the old syntax form of `onClickOutside(componentClass, [options])` as you suggested and synced it with your current HEAD so it can be merged. Sorry it took so long...

Pomax:
no worries - I'm traveling at the moment but will be back home later in the weekend, so I'll try to get this merged in before the new week

README.md
```diff
@@ -28,8 +28,27 @@ var MyComponent = onClickOutside(React.createClass({
 }));
 
 ```
+Note that if you try to wrap a React component class without `handleClickOutside(evt)` handler like this, the HOC will throw an error.
```
Pomax: "class without" => "class without a"

README.md
```diff
@@ -28,8 +28,27 @@ var MyComponent = onClickOutside(React.createClass({
...
+If you want to execute method different than `handleClickOutside(evt)`, you can specify it as a second parameter to `onClickOutside` like this:
```
Pomax: "In order to use a custom event handler, you can specify the function to be used in the HOC options objects, passed as second parameter:"

README.md
```diff
@@ -28,8 +28,27 @@ var MyComponent = onClickOutside(React.createClass({
...
+var onClickOutside = require('react-onclickoutside');
+
+// create a new component, wrapped by this onclickoutside HOC:
+var MyComponent = onClickOutside(React.createClass({
+  ...,
+  myClickOutsideHandler: function(evt) {
+    // ...handling code goes here...
+  },
+  ...
+}), function (instance) {
```
Pomax: looking at this some more, this should probably simply be part of an options object instead, to be more in line with virtually all modern JS libraries.
```
var Thing = onClickOutside(component, {
  handleClickOutside: component.functionNameForThat
});
```

README.md
```diff
@@ -28,8 +28,27 @@ var MyComponent = onClickOutside(React.createClass({
...
+    instance.myClickOutsideHandler
+});
+
+```
+
+Note that if you try to wrap a React component class while specifying a custom handler and the component does not implement it, the HOC will throw an error.
```
Pomax: we probably want to explicitly say the error will happen at compile time

Pomax:
Back from the US - looking at the best way to use this, I left a few more comments (some to improve the README, but one is fairly important because it's about how to pass config options, which is relevant to a second PR that is currently open, which deals with passing in more config options, so changing the 2nd argument to an options object gives a universally extensible config, rather than something that has to be a function, necessitating a third argument for additional config options)

README.md
```diff
@@ -42,13 +43,15 @@ var MyComponent = onClickOutside(React.createClass({
     // ...handling code goes here...
   },
   ...
-}), function (instance) {
-    instance.myClickOutsideHandler
+}), {
+  handleClickOutside: function(instance) {
```
no23reason: I could not find a way to use the `handleClickOutside: Component.myClickOutsideHandler`, so I stuck with the accessor function notation.

no23reason:
@Pomax thanks for the feedback, I pushed a commit that includes the changes you proposed. Sorry for the mess up with the config parameter...

Pomax:
merge'd

PR #119, opened 2016-08-20
fallback to call handleClickOutside from props, fix #98

### LINKED ISSUE #98: Support for stateless components
I tried to use this lib with [recompose](https://github.com/acdlite/recompose/blob/master/docs/API.md#setproptypes):

```js
const controller = compose(
    withHandlers({
        handleClickOutside: (props) => (event) => null // do something
    }),
    onClickOutside
);

const component = controller((props) => <div>...</div>);
```

Unfortunately, onClickOutside HOC expects component with `handleClickOutside` method defined in component itself, so it doesn't work with stateless components.

Right now I'm using simple component using existing API which I use a root component in my stateless one:

```js
@onClickOutside
class OutsideClickHandler extends Component {
  handleClickOutside() {
    this.props.handleClickOutside();
  }

  render() {
    return this.props.children;
  }
}
```

Question is: Is it worth to include such a component in this library? Is there a better way how to support stateless components or take handler function from props?

kevin940726:
Quick fix for #98, simply fallback to call `props.handleClickOutside` instead.
Also there is a simple test.

### DIFF
--- a/index.js
+++ b/index.js
@@ -107,10 +107,13 @@
               throw new Error("Component lacks a function for processing outside click events specified by the handleClickOutside config option.");
             }
           } else {
-              if(typeof instance.handleClickOutside !== "function") {
-                throw new Error("Component lacks a handleClickOutside(event) function for processing outside click events.");
+            if(typeof instance.handleClickOutside === "function") {
+              clickOutsideHandler = instance.handleClickOutside;
+            } else if(typeof instance.props.handleClickOutside === "function") {
+              clickOutsideHandler = instance.props.handleClickOutside;
+						} else {
+              throw new Error("Component lacks a handleClickOutside(event) function for processing outside click events.");
             }
-            clickOutsideHandler = instance.handleClickOutside;
           }
 
           var fn = this.__outsideClickHandler = generateOutsideCheck(
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@
  * A higher-order-component for handling onClickOutside for React components.
  */
 (function(root) {
-  "use strict";
+  'use strict';
 
   // administrative
   var registeredComponents = [];
@@ -57,7 +57,7 @@
       // outside", as it cannot be known whether it was outside.
       if(current !== document) return;
       eventHandler(evt);
-    }
+    };
   };
 
 
@@ -91,7 +91,7 @@
         },
 
         // this is given meaning in componentDidMount
-        __outsideClickHandler: function(evt) {},
+        __outsideClickHandler: function() {},
 
         /**
          * Add click listeners to the current document,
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -56,7 +56,12 @@
       // If element is in a detached DOM, consider it "not clicked
       // outside", as it cannot be known whether it was outside.
       if(current !== document) return;
-      componentInstance.handleClickOutside.bind(componentInstance, evt)();
+
```
Pomax: can you please fix the indentation here? (code uses 2-space indentation)

test/test.js
```diff
@@ -56,4 +56,33 @@ describe('onclickoutside hoc', function() {
       assert(e, "component was not wrapped");
     }
   });
+
```
Pomax: please fix the indentation here, too

kevin940726:
Sorry for not noticing that! 🙇 
Maybe an eslint and editorconfig file would help us with that by catching the styling issue automatically.
Please let me know if there still have any issue, thanks!

Pomax:
yeah, it's been a pretty lean codebase without any builds, testing only recently got added, but a a `jscs` task is probably a good idea. let me file that!

Also due to a previous PR landing github is showing this PR as needing a rebase, that's entirely my doing, so with apologies: could you also rebase this PR so we can merge it in?

kevin940726:
Sure! though [jscs is actually merged with eslint](https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2#.3je21spti), it may be a good idea to just use [eslint](http://eslint.org/) instead. I am glad to help with that, let me know if you need some help! 😊 

I have merged and solved the conflict, thanks!

Pomax:
yeah, I filed the issue as just an "add eslint" issue. Thanks for the rebase!

test/test.js
```diff
@@ -88,14 +87,43 @@ describe('onclickoutside hoc', function() {
     });
 
     var element = React.createElement(WrappedWithCustomHandler);
-    assert(element, "element can be created");
+		assert(element, "element can be created");
```
Pomax: mystery indentation?

test/test.js
```diff
@@ -88,14 +87,43 @@ describe('onclickoutside hoc', function() {
...
     var component = TestUtils.renderIntoDocument(element);
     assert(component, "component renders correctly");
     document.dispatchEvent(new Event('mousedown'));
     var instance = component.getInstance();
     assert(instance.state.clickOutsideHandled, "clickOutsideHandled got flipped");
   });
 
+	it('should fallback to call component.props.handleClickOutside when no component.handleClickOutside is defined', function() {
```
Pomax: looks like some of the tabs slipped back in O_O

kevin940726:
omg, I am so sorry! We really need a linter :sweat_smile: !
Let me fix this with a basic working eslint file in the next commit, thanks.

kevin940726:
#120 
The eslint file only contains basic rules such as...
* indent style 2 spaces
* single quotes string
* always required semicolon
* and no un-used variables
I don't know if that fit your needs but feel free to adjust or add any rules you like 😊 

I also add a `husky` package to automatically lint the directory whenever we commit. We can also run the test per commit or per push if you prefer.

Pomax:
merged the ESlint and fallback code (https://github.com/Pomax/react-onclickoutside/pull/124) separately, as this PR had weird webpack errors due to the husky and assert depdencies, neither of which were necessary (karma does its own interpretation magic so that `asset` is simply available, it doesn't need a special requirement - it just needed to be cleared as global in the eslint config)

PR #124, opened 2016-09-05
fallback to call handleClickOutside from props

### LINKED ISSUE #98: Support for stateless components

Pomax:
fixes #98

### DIFF
--- a/index.js
+++ b/index.js
@@ -107,12 +107,13 @@
             if(typeof clickOutsideHandler !== 'function') {
               throw new Error('Component lacks a function for processing outside click events specified by the handleClickOutside config option.');
             }
-          } else {
-            if(typeof instance.handleClickOutside !== 'function') {
-              throw new Error('Component lacks a handleClickOutside(event) function for processing outside click events.');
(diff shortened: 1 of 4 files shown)


Pomax:
v5.5.0 released

Issue #87, opened 2016-04-26
React/ES6 handleClickOutside not bound


calvinnwq:
Before v5.0.0, when you used the HOC, handleClickOutside was bound. Now it seems I have to bind it in my constructor, otherwise I cannot access `this.props` in my event handler.

```
constructor() {
  super();
  this.handleClickOutside = this.handleClickOutside.bind(this);
}
```

Should this be the case?

Pomax:
Hm, it shouldn't - if you change https://github.com/Pomax/react-onclickoutside/blob/master/index.js#L108 to read `instance.handleClickOutside.bind(instance)` does the problem go away? If so, that should be a fairly quick fix (and can be spun off into a release relatively easily)

calvinnwq:
Updating line 108 to `instance.handleClickOutside.bind(instance)` fixes it.

PR #125, opened 2016-09-06
Fixes regression of #88

RichGuk:
I think #114 introduced a regression of #87/#88.

This code should fix it (at least it does for me 👍), unless I'm not understanding something correctly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -108,12 +108,12 @@
               throw new Error('Component lacks a function for processing outside click events specified by the handleClickOutside config option.');
             }
           } else if(typeof instance.handleClickOutside === 'function') {
-            clickOutsideHandler = instance.handleClickOutside;
+            clickOutsideHandler = instance.handleClickOutside.bind(instance);
           } else if(typeof instance.props.handleClickOutside === 'function') {
-            clickOutsideHandler = instance.props.handleClickOutside;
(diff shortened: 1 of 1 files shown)


jbblanchet:
I can confirm that this fixes the problem we had with version 5.5.

semanser:
This solution works perfect for me.

jbblanchet:
I started by only implementing the tests for your PR, but while implementing the tests I discovered that we shouldn't always re-bind, because react.createClass is taking care of it, so I updated the code to check how the component was created. Sorry to steal your PR :sweat:

Pomax:
Does this reintroduce the whole "React complaining about `bind()` calls"?

RichGuk:
@jbblanchet  That's fine, I didn't test with createClass.

@Pomax I did wonder that. Which commit actually fixed that?  9d412ca5f832ff813faf0b427a2942400941633d ? Few merges/reverts, was confusing.

Pomax:
Eh, I'll be honest - if this fixes a real bug, then reintroducing a warning is fine. But, if you'd like to help figure out how to now *remove* that warning again, that'd be fantastic.

Pomax:
v5.6.0 just got released after merging in @jbblanchet's PR, does that solve the problem this PR also tried to solve?

RichGuk:
It does thank you @Pomax and @jbblanchet  :)

PR #126, opened 2016-09-06
add test for binding issues

jbblanchet:
This was a bit more complex than anticipated, because I had to add babel to reproduce the issue, but this test will fail if the regression fixed by #125 comes back.

This also brings to the forefront the difference between components created by inheritance, and components created by createClass. ~~The re-binding will give an error on the latter. Not sure what a long term solution is.~~ This can be fixed by looking at the prototype of the element.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,9 +30,11 @@
     "object-assign": "^4.0.1"
   },
   "devDependencies": {
+    "babel-preset-es2015": "^6.14.0",
     "chai": "^3.5.0",
     "eslint": "^3.4.0",
     "karma": "^0.13.22",
+    "karma-babel-preprocessor": "^6.0.1",
     "karma-chai": "^0.1.0",
     "karma-mocha": "^0.2.2",
     "karma-phantomjs-launcher": "^1.0.0",
--- a/test/.eslintrc.json
+++ b/test/.eslintrc.json
@@ -1,0 +1,5 @@
+{
+  "parserOptions": {
+    "ecmaVersion": 6
+  }
+}
--- a/test/test.js
+++ b/test/test.js
@@ -66,41 +66,160 @@ describe('onclickoutside hoc', function() {
       assert(e, 'component was not wrapped');
     }
   });
+  
 
+  describe('with instance method', function() {
+    it('and class inheritance, should call the specified handler when clicking the document', function() {
+      class Component extends React.Component {
+        constructor(...args) {
+          super(...args);
+          this.state = {
+            clickOutsideHandled: false
+          };
+        }
 
-  it('should call the specified handler when clicking the document', function() {
-    class Component extends React.Component {
-      constructor(...args) {
-        super(...args);
-        this.state = {
-          clickOutsideHandled: false
-        };
+        handleClickOutside (event) {
+          if (event === undefined) {
+            throw new Error('event cannot be undefined');
+          }
+
+          this.setState({
+            clickOutsideHandled: true
+          });
+        }
+
+        render() {
+          return React.createElement('div');
+        }
       }
 
-      handleClickOutside (event) {
+      var WrappedWithCustomHandler = wrapComponent(Component);
+
+      var element = React.createElement(WrappedWithCustomHandler);
+      assert(element, 'element can be created');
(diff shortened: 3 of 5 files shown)


jbblanchet:
This one should be ready to merge, but I'd still like to add tests to check all the possibilities (createClass vs. React.Component vs. instance method vs. props etc.). I'll either update this PR if not merged yet or create a new one when I have time to add those later today.

package.json
```diff
@@ -30,9 +30,11 @@
     "object-assign": "^4.0.1"
   },
   "devDependencies": {
+    "babel-preset-es2015": "^6.14.0",
```
Pomax: why do we need these? `isPrototypeOf` has been in JavaScript since ECMAScript 3rd edition, so no babelification should be necessary here?

test/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "parserOptions": {
+    "ecmaVersion": 6
```
Pomax: no ES6 is used in the code, so this should not be necessary

test/test.js
```diff
@@ -69,33 +69,30 @@ describe('onclickoutside hoc', function() {
 
 
   it('should call the specified handler when clicking the document', function() {
-    var Component = React.createClass({
-      getInitialState: function() {
-        return {
+    class Component extends React.Component {
```
Pomax: why not use a proper stateless component, which is just a `function(props) { return UICodeGoesHere }`?

test/.eslintrc.json
jbblanchet: @Pomax we're using classes and inheritance to be able to reproduce the issue.

jbblanchet: @Pomax But only in the tests.

Pomax: left a note there, too. I don't see why the ES6 rewrite was necessary. Stateless components are simply functions with signature `function(props) { return whateveryouneedrendered }`.

test/test.js
jbblanchet: @Pomax A stateless component might be sufficient to generate the error, but I do think that testing the 3 ways to create a component (stateful component with inheritance, createClass and stateless function) will give you the best chance to catch these errors early next time.

I should be able to add those tests by the end of the day. If you prefer that I only add tests for stateless vs createClass, let me know.

jbblanchet:
@Pomax since it seems that there still should be a discussion on the tests, and that the actual fix should be merged as soon as possible (because 5.5 introduced a breaking change over 5.3), do you want me to create a separate PR for the fix so you can release it ASAP, then update this one with the tests?

test/test.js
jbblanchet: Still thinking about your comment, and stateless functions won't be re-bound here, because they do not inherit from React.Component. I'm not sure if it makes sense to support instance method on stateless functions, but if so, the test should be reversed to check if createClass was used. The (very ugly) way to do that appears to be:

```javascript
var ReactClassComponentPrototype = (Object.getPrototypeOf(Object.getPrototypeOf(new (React.createClass({ render: function () {} }))())))
```

Source: https://discuss.reactjs.org/t/how-to-determine-if-js-object-is-react-component/2825/5

jbblanchet: But (still thinking out loud) I don't think it makes sense to auto bind on the instance on stateless functions. The whole point of those is to be... stateless. So the logic here would be:
- If it has been created by inheritance: re-bind.
- If it has been created with createClass: binding already taken care of, do nothing.
- If it it a stateless method: binding shouldn't be necessary, do nothing.

But to test that logic, I do need to use es6 classes, unless I try and mock the inheritance, but that seems risky and more complex.

jbblanchet:
Added some more tests to cover the use case I could think of. Components are now created by inheritance (with instance method and prop), with a createClass function (with instance method and prop) and with a stateless function (with prop). That last test is not working, because stateless functions do not support refs, so I skipped it for now. It also made me realize that re-binding was not necessary on props, since props should not be bound to the instance, so I updated the code.

Pomax:
new code yields a pretty readable `npm test` log so I'm happy to merge this in.

Pomax:
v5.6.0 released

PR #128, opened 2016-09-09
Remove ref for stateless components

### LINKED ISSUE #127: Stateless component causes errors trying to read refs
Related: https://github.com/Pomax/react-onclickoutside/issues/98

Stateless component support was recently added, but seems to be broken. I tested this with a simple react app (I mostly wanted to try out Create React App): https://github.com/davishmcclurg/react-onclickoutside-issue

The stateless component is defined here: https://github.com/davishmcclurg/react-onclickoutside-issue/blob/master/src/App.js

This is what I see in the console:

> Warning: Stateless function components cannot be given refs (See ref "instance" in StatelessComponent created by OnClickOutside(Component)). Attempts to access this ref will fail.

And:

> Uncaught TypeError: Cannot read property 'handleClickOutside' of null

davishmcclurg:
Stateless components don't support refs, so this
uses the props from the wrapper component instead.
`isReactComponent` isn't documented, but seems to
work for both ES6 classes and components created
with `React.createClass`.

Fixes #127

### DIFF
--- a/index.js
+++ b/index.js
@@ -88,7 +88,7 @@
          * Access the wrapped Component's instance.
          */
         getInstance: function() {
-          return this.refs.instance;
+          return Component.prototype.isReactComponent ? this.refs.instance : this;
         },
 
         // this is given meaning in componentDidMount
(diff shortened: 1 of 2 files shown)


Pomax:
nice, tests pass so in this goed!

Pomax:
v5.7.0 released

Issue #136, opened 2016-09-26
Fails completely if rendering into an environment without a DOM


dariusriggins:
I'm adding snapshot testing from storyshots (conceptually from Jest) which is failing due to not being in an environment with a DOM. In my instance it's failing because `ReactDOM.findDOMNode is not a function` inside of `componentDidMount`.

While I don't expect this to _work_ in a non-DOM environment I would expect that it wouldn't fail and break my tests if in an environment it doesn't understand.

Pomax:
why not? As a very specifically DOM-grounded hoc, that's one thing I absolutely would expect. PRs welcome to fix that, but expecting it to work without a DOM from the outset doesn't jive =)

dariusriggins:
Yeah it's not ideal, but right now my only way to have tests is to conditionally wrap my component in the hoc if I'm running tests, which feels kinda hacky. 

And again, I don't expect it to functionally work, I would just like some manner to prevent it from a hard failure in that scenario. I'd imagine shallow rendering with testutils or Enzyme would also encounter the same thing.

dariusriggins:
Cool, if you're fine with a try/catch or check for it then I'm happy to put together a PR.

Pomax:
yeah that's fine by me, or even an `if (!ReactDOM.findDOMNode) { return early  }`. As long as it does not impair the code from doing what it needs to do when there is a dom to work with, I'm open to solutions.

Pomax:
if that's the only way we can be sure we don't have a DOM, then that sounds fine - I'm a proponent of "solve a problem, then refine the approach when someone runs into edge cases with the solution"

Pomax:
v5.7.1 released

PR #138, opened 2016-09-28
Dont fail if rendering in an environment with no DOM

dariusriggins:
https://github.com/Pomax/react-onclickoutside/issues/136

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,6 +99,13 @@
          * linked to this component's state.
          */
         componentDidMount: function() {
+          // If we are in an environment without a DOM such 
+          // as shallow rendering or snapshots then we exit 
+          // early to prevent any unhandled errors being thrown.
+          if (!ReactDOM.findDOMNode){
+            return;
+          }
+
           var instance = this.getInstance();
           var clickOutsideHandler;
 
--- a/index.js
+++ b/index.js
@@ -67,7 +67,7 @@
    * bootstrapping code to yield an instance of the
    * onClickOutsideHOC function defined inside setupHOC().
    */
-  function setupHOC(root, React, ReactDOM) {
+  function setupHOC(root, React, ReactDOM, ExecutionEnvironment) {
 
     // The actual Component-wrapping HOC:
     return function onClickOutsideHOC(Component, config) {
@@ -102,7 +102,7 @@
           // If we are in an environment without a DOM such 
           // as shallow rendering or snapshots then we exit 
           // early to prevent any unhandled errors being thrown.
-          if (!ReactDOM.findDOMNode){
+          if (!ExecutionEnvironment.canUseDOM){
             return;
           }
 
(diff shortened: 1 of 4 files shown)


dariusriggins:
Wasn't super happy to not be able to use the existing karma tests, but until that bug is fixed there's no other way to render to a different environment that I could find.

index.js
```diff
@@ -99,6 +99,13 @@
          * linked to this component's state.
          */
         componentDidMount: function() {
+          // If we are in an environment without a DOM such 
+          // as shallow rendering or snapshots then we exit 
+          // early to prevent any unhandled errors being thrown.
+          if (!ReactDOM.findDOMNode){
```
Pomax: curious if this is the most authoritative way to determine we have no DOM at all (I would imagine that if ReactDOM is required in correctly, then the findDomNode function at least exists, even if it won't actually work).

dariusriggins: I'm not entirely sure if it's the most authoritative route. I did find https://www.npmjs.com/package/exenv which is a port of what React uses to determine the environment that could be used to do the detection instead of this.

I'm looking through ReactDOM and findDOMNode but it's not yet immediately clear why it's not being required in at all in the non-DOM case. If I look at ReactDOM in this state the only method available off of it is `render`, neither `findDOMNode`, nor `version` or any of the others are available.

Let me switch this out with the environment check and see if it's still detecting it properly.

dariusriggins:
I've updated it with the test for ExecutionEnvironment, although I do wonder if the license is something that's OK to bring in as it's BSD. It also seems it's a relatively simple check with the definition being this: https://github.com/JedWatson/exenv/blob/master/index.js#L11

```javascript
	var canUseDOM = !!(
		typeof window !== 'undefined' &&
		window.document &&
		window.document.createElement
	);
```

Pomax:
I am no lawyer, but this does not strike me as the kind of code anyone can lay claim on, it's a straightforward "do we have window? k, do we have document? k, do we have this function in document that we need?" check (thousands of client+serer libraries employ the same checks).

We don't need the double negation though, JS coercion means that `if (canUseDOM) { .... }` will work just fine. That said, we shouldn't need to check for `window`, the `document` is also a global variable in "for real" DOM environments (browser but also things like jsdom), so we could just do:
```
var canUseDOM = (typeof document !== "undefined" && document.createElement);
```

dariusriggins:
Yeah, I'll remove that lib as a dependency and inline the check for the DOM. That seems reasonable and passes the tests.

I hear what you're saying on "for real" DOM environments, that lib from what I was reading copied it from react internal code. I'm not sure the purpose either, document is not a thing in web workers either, so perhaps we go with this for now until as you said we run into an edge case.

dariusriggins:
Since there's only one use I just inlined it, let me know if you'd prefer I make that more of a global thing inside `index.js`.

dariusriggins:
@Pomax Is there any changes you'd like to see on this PR? Just want to make sure you're not waiting on me.

Pomax:
I think we can probably merge this in. I'm currently traveling so I've not been online all that much.

dariusriggins:
Thanks for getting this in so quickly, and for making this, we find a lot of really great use and it almost always just works. Cheers!

PR #143, opened 2016-10-21
IE11 - handle elements without classList

### LINKED ISSUE #142: IE11 - SVGSVGElement missing classList
We've hit a NPE on IE11 with inline SVG:

     SCRIPT5007: Unable to get property 'contains' of undefined or null reference

It turned out `isNodeFound` is testing CSS classes on a `SVGSVGElement` which, as per Microsoft documentation, doesn't have a `classList` property: 

- https://msdn.microsoft.com/en-us/library/ff972122(v=vs.85).aspx

elsassph:
Fixes #142

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@
     if (current.correspondingElement) {
       return current.correspondingElement.classList.contains(ignoreClass);
     }
-    return current.classList.contains(ignoreClass);
+    return current.classList && current.classList.contains(ignoreClass);
   };
 
   /**


Pomax:
https://github.com/Pomax/react-onclickoutside#ie-does-not-support-classlist-for-svg-elements

elsassph:
Oh come on, this PR is NOT to make the lib work with IE's SVG but to ignore those. I don't need/want to shim it, there's just bits of SVG in the tree.

Pomax:
You don't want to shim it, you just rely on code that isn't supported by IE older than EDGE, so just shim it anyway. IE-specific fixes will not be accepted into this codebase.

PR #147, opened 2017-01-19
Remove unused dependency from package.json

suburbanmtman:
The object-assign dependency was not being used and can thus be safely removed.


Pomax:
nice catch

suburbanmtman:
Great, thanks!

Issue #130, opened 2016-09-12
Scrollbar click and/or drag causes close to fire


Rodeoclash:
This is a tricky one because technically it could be the correct behaviour.

If the user clicks and drags a top level scrollbar in the browser chrome, it will fire the on close event. Here's an example of a floating navigation bar with content underneath it. You can see me dragging it, then opening the popup and trying to do the same thing. The popup will fire the `handleClickOutside` and close on mousedown on the scrollbar.

https://www.youtube.com/watch?v=rVA6F9hFgM4

Pomax:
Is this the built-in scrollbar?

Pomax:
Wow, that's pretty amazing. The document scrollbar is considered part of the `<html>` element, so that's a neat problem to deal with.

Pomax:
We can probably check, at the `html` element, whether the event's `x` coordinate is larger than `document.documentElement.clientWidth`. If so, despite being a click "on the html element", that will have been a click on the associated scrollbar, not the html element's content.

Pomax:
This seems to also yield the correct result for Firefox.

Rodeoclash:
Checking the document dimensions is pretty much the conclusion I came to as well. Ugly though. And I'm not convinced the existing behaviour is actually incorrect, just unexpected.

Rodeoclash:
Do you want me to try and put together a PR for this?

Rodeoclash:
Even stranger, click and dragging on internal scrollbars, e.g. and `overflow: scroll;` DOES NOT trigger the close! Fun!

Pomax:
yeah the `html` level scrollbars, as far as I can tell, are very special this way, which is why document dimensions actually work (it's the only scrollbar for which this trick should work). 

If you want to try your hand at a PR, that'd be fantastic! It might also need a test though, not sure a unit test will cover this particular one... we might need a manual in-browser test.

Rodeoclash:
Sounds good. I'll wait until our PM's scream loudly enough here about it to allocate some time to fixing it :smile:

Pomax:
closing for now, but feel free to reopen or refile if this becomes a problem again.

PR #149, opened 2017-02-07
Add option to exclude scrollbar from outside check

phwebi:
This PR addresses https://github.com/Pomax/react-onclickoutside/issues/130 by allowing an option to exclude the browser scrollbar from the click outside check.

The default behavior would be the same as current behavior.

### DIFF
--- a/index.js
+++ b/index.js
@@ -52,10 +52,17 @@
   };
 
   /**
+   * Check if the browser scrollbar was clicked
+   */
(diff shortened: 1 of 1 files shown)


Pomax:
Nice! In it goes.

Pomax:
v5.9.0 published to npm

phwebi:
Woohoo :)

PR #153, opened 2017-03-01
Reset handler on update

jtadmor:
This PR addresses the scenario in which the HOC wraps a component that does something like this:

```
const MyMenu = ({ items, asDiv }) => (asDiv? <div>{items}</div> : <ul>{items}</ul>)
```

In that case, if `asDiv` starts as `true` but changes to `false` while the component is rendered, we end up having the click handler listening on the wrong node, and any click inside the `<ul>` is registered as outside.

This PR checks whether the DOM node rendered is the same after each update, and, if not, it updates the handler to the new component.

This also has the advantage of printing warnings if a component switches to returning `null` from its `render()` even though it originally returned a valid DOM element when first rendering.

### DIFF
--- a/index.js
+++ b/index.js
@@ -162,14 +162,19 @@
               'for that decision.'
             ].join(' '));
           }
-          
+
           if (this.__node === componentNode) {
             /*
-              If the node hasnt changed after an update, then return
+              If the node hasnt changed after an update, then return.
             */
             return;
+          } else {
+            /*
+              Disable existing click handlers because we will set new ones.
+            */
+            this.disableOnClickOutside();
           }
-          
+
           this.__node = componentNode;
 
           var fn = this.__outsideClickHandler = generateOutsideCheck(
@@ -181,12 +186,12 @@
             this.props.preventDefault || false,
             this.props.stopPropagation || false
           );
-          
+
           if (!this.__pos) {
             this.__pos = registeredComponents.length;
             registeredComponents.push(this);
           }
-         
+
           handlers[this.__pos] = fn;
 
           // If there is a truthy disableOnClickOutside property for this
@@ -195,14 +200,14 @@
             this.enableOnClickOutside();
           }
         },
-        
+
         componentDidMount: function() {
           this.__setOutsideClickHandler();
         },
-        
+
         componentDidUpdate: function(prevProps) {
           this.__setOutsideClickHandler();
-          
+
           if (this.props.disableOnClickOutside && !prevProps.disableOnClickOutside) {
             this.disableOnClickOutside();
           } else if (!this.props.disableOnClickOutside && prevProps.disableOnClickOutside) {
--- a/index.js
+++ b/index.js
@@ -116,7 +116,7 @@
          * Add click listeners to the current document,
          * linked to this component's state.
          */
-        componentDidMount: function() {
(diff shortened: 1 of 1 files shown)


jtadmor:
All the tests seem to be passing, but I recognize this this PR carries (at least) two potential risks:

1. Performance hit by checking the DOM node with every component update.

2. The check for adding / removing the click handlers used to be in `componentWillReceiveProps` and is now in `componentDidUpdate`. There might be a brief timing window where a handler might be active / not-active where that wasnt the case before. That should be pretty negligible but I don't wanna break anything that relied on it.

_Potential Solution:_ Offer a new prop / config option, like `recheckOnUpdate` for when the component knows it might render different roots. If this config is true, the new check happens in `componentDidUpdate`, if not, the HOC works exactly as it does today.

This allows people to opt into the new behavior. Happy to put that code in there.

Pomax:
I'm not sure I'm up for a fix that addresses a scenario where "asDiv starts as true but changes to false while the component is rendered". If that value changes, and that change should impact your component UI, then that value is a component state value plain and simple. As such, you should be accessing it inside the handler by calling `this.state.asDiv`.

I'm going to need a bit of convincing that you're not trying to change how the HOC works to make it compatible with code that is itself highly questionable =/

jtadmor:
That's totally fair. There's at least one open source project out there that this is causing a bug in (https://github.com/YouCanBookMe/react-datetime), because it follows a pattern like this:

```
class MyList extends Component {
  render() {
    if (this.state.view === 'time') {
      return <TimePicker />
    } else {
      return <DatePicker />
    }
}
```

(Is this what you mean by having it as a state value?)

Any component like that, potentially rendering an entirely new root, wont work. 

Totally understandable if you'd rather leave it out of this library, components like the above could separate wrap each individual component (`TimePicker` and `DatePicker`) or if folks really need it they can fork.

Pomax:
The code you show is a perfectly correct use of React, where `render` generates different UI based on a component's state. However, the code you showed as example of what react-onclickoutside was unable to copy with gives a function handler 
```
const MyMenu = ({ items, asDiv }) => (asDiv? <div>{items}</div> : <ul>{items}</ul>)
```
where MyMenu is a function that returns either a div, or a list, based on the *function argument* `asDiv`, where in stead it should always be looking at the component state:
```
const MyMenu = (items) => (this.state.asDiv? <div>{items}</div> : <ul>{items}</ul>);
```
so that every time it is invoked, the component's state is invoked to make sure the correct data is generated. However, without the code around it it's hard to say where else the code may need restructuring to do things in a "Reacty" way. 

Can you show a minimal example of a full component that uses the construction you're trying to deal with?

jtadmor:
Sorry, I was just using the functional stateless component to demonstrate the point quickly.

The `MyList` class that I used above would have this bug, it is the minimal representation of what is happening in the datepicker lib.

Pomax:
I'd still like to understand this problem as a problem, and it kind of feels like I need a PoC to understand this use case, because a PoC will let me see whether or not something weird is going on in the code, or whether this is a legitimate use-case that needs addressing in the HOC instead

Pomax:
I'm going to close this for now simply because there hasn't been any activity for 2 weeks. If I should reopen it, let me know, but then the use-case should be made clear before review continues

PR #154, opened 2017-03-07
Adding ES6 Usage Example to Readme.md

jtulk:
I went digging in the issues to find an example for ES6 React Classes. Thought I try to slip the example into the documentation.


Pomax:
try and succeed, you mean.

PR #157, opened 2017-03-12
fix excludeScrollbar not working for horizontal scroll bar

yormo:
exclude horizontal Scrollbar

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,7 +55,7 @@
    * Check if the browser scrollbar was clicked
    */
   var clickedScrollbar = function(evt) {
-    return document.documentElement.clientWidth <= evt.clientX;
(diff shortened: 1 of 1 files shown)


Pomax:
I'm a little confused: if this is a fix for the horizontal scrollbar, why does the code introduce a *height* check, which would incur a vertical scrollbar, if any?

rafeememon:
@pomax, I think this is because a horizontal scroll bar would reduce the height of a page since it appears along the bottom.  The existing vertical scroll bar check is analogous in that it checks for a reduced width.

yormo:
@Pomax When horizontal scrollbar appear, clientHeight will become smaller, Through ‘clientY’ to check whether the event on horizontal scrollbar, This method is similar to your check vertical scrollbar.

Pomax:
I see. Would you be willing to write a mocha test (./test/test.js) to verify this is doing the right thing?

PR #158, opened 2017-03-16
excludeScrollbar can now be set as default for every component instance

lorenzos:
I added the ability to use `excludeScrollbar` for every instance of the wrapper component, using an option in the already existing `config` object that can be passed as second parameter to the `onClickOutside()` wrapper.

This is an opt-in that do not change the current behavior, and that can still be overriden by setting the `excludeScrollbar` prop per-instance as usual.

Implementation is quite neat: I simply used `getDefaultProps()` to set the `excludeScrollbar` prop, so the the feature is still working exactly as it was before.

### DIFF
--- a/index.js
+++ b/index.js
@@ -174,7 +174,7 @@
             instance,
             clickOutsideHandler,
             this.props.outsideClickIgnoreClass || IGNORE_CLASS,
-            this.props.excludeScrollbar,
+            this.props.excludeScrollbar, // fallback not needed, prop always exists because of getDefaultProps
             this.props.preventDefault || false,
             this.props.stopPropagation || false
           );
--- a/index.js
+++ b/index.js
@@ -112,6 +112,12 @@
         // this is given meaning in componentDidMount
         __outsideClickHandler: function() {},
 
+        getDefaultProps: function() {
+          return {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -112,6 +112,12 @@
         // this is given meaning in componentDidMount
         __outsideClickHandler: function() {},
 
+        getDefaultProps: function() {
+          return {
+            excludeScrollbar: config && config.excludeScrollbar
```
Pomax: with this, https://github.com/Pomax/react-onclickoutside/blob/master/index.js#L171 will always exist, so we shouldn't need that `|| false` fallback anymore (with a comment `// pulls from defaultProps`

lorenzos: Done.

index.js
```diff
@@ -174,7 +174,7 @@
             instance,
             clickOutsideHandler,
             this.props.outsideClickIgnoreClass || IGNORE_CLASS,
-            this.props.excludeScrollbar || false,
+            this.props.excludeScrollbar,
```
Pomax: can you also add the comment that explains why this has no fallback? good for future code-readers to have

Pomax:
awesome, time to merge this in. Thanks!

PR #164, opened 2017-04-08
Migrated deprecated React.createClass

Andarist:
I think we should migrate to the es6 `class` syntax and stop using `React.createClass`, but that would be another step. For now just created this PR, so people stop having deprecation warnings after migration to React v15.5. 

If you feel the same about `class`, I'll create an issue about it, so the matter won't get forgotten and neglected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,10 @@ $> npm install react-onclickoutside --save
 ```js
 // load the HOC:
 var onClickOutside = require('react-onclickoutside');
+var createReactClass = require('create-react-class');
 
 // create a new component, wrapped by this onclickoutside HOC:
-var MyComponent = onClickOutside(React.createClass({
+var MyComponent = onClickOutside(createReactClass({
   ...,
   handleClickOutside: function(evt) {
     // ...handling code goes here...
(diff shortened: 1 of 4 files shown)


Pomax:
awesome, I was going to put this in over the weekend after the announcement but you beat me to it.

I would like to switch over to `class` code only once React 16 comes out, with a major version switchover to v6.x.x that aligns with the React 16 APIs.

Pomax:
Hm, this code does need some follow up, though - I forgot to run `npm test` before merging and this completely fails to run in the browser because the `createReactClass` function cannot be found anywhere in the 15.5 react-with-addons library.

Pomax:
followup pushed in https://github.com/Pomax/react-onclickoutside/commit/5b023e90b09ea148285296c31041e35d7aa5b8dd

Pomax:
v5.11.0 published and `README.md` updated to mention the React 15.5 / v.5.11.x pairing.

Andarist:
Thanks for fixing issues. 

> createReactClass function cannot be found anywhere in the 15.5 react-with-addons library.

According to the [anouncement - "We're also discontinuing support for the react-with-addons UMD build. It will be removed in React 16."](https://facebook.github.io/react/blog/). So I believe this is intentionally missing from there.

Pomax:
I'm sure it is, but it does mean this PR actually needed a fair bit more work before it was good to merge - that's on me, I need to hook up a travis CI pr hook...

PR #166, opened 2017-04-10
Add an option to create non-passive event listeners

mrblackus:
Passive event listeners does not block on `touch` and `wheel` events because they will not wait to see if `preventDefault()` is called. It allows to have better performance on scrolling ([more details here](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md) and [here](https://developers.google.com/web/updates/2016/06/passive-event-listeners)).

All event listeners are now created on passive mode on Chrome [since version 56](https://www.chromestatus.com/features/5093566007214080), and it might be default on other browser in the future.

To explicitely make the created event listeners not passive, allowing us to call `preventDefault` on them, add `{passiveListener: false}` as an option to the HOC.

```js
export default onClickOutside(Component, { passiveListener: false });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,15 @@ Technically this HOC lets you pass in `preventDefault={true/false}` and `stopPro
 
 Each component adds new event listeners to the document, which may or may not cause as many event triggers as there are event listening bindings. In the test file found in `./test/browser/index.html`, the coded uses `stopPropagation={true}` but sibling events still make it to "parents".
 
+## Handling passive events
+
+Passive event listeners does not block on touch and wheel events because they will not wait to see if `preventDefault()` is called. It allows to have better performance on scrolling ([more details here](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md) and [here](https://developers.google.com/web/updates/2016/06/passive-event-listeners)).
+To explicitely make the created event listeners not passive, allowing us to call `preventDefault` on them, add `{passiveListener: false}` as an option to the HOC.
+
+```js
+export default onClickOutside(Component, { passiveListener: false });
+```
+
 ## Marking elements as "skip over this one" during the event loop
 
 If you want the HOC to ignore certain elements, you can tell the HOC which CSS class name it should use for this purposes. If you want explicit control over the class name, use `outsideClickIgnoreClass={some string}` as component property, or if you don't, the default string used is `ignore-react-onclickoutside`.
--- a/index.js
+++ b/index.js
@@ -115,7 +115,7 @@
         getDefaultProps: function() {
           return {
             excludeScrollbar: config && config.excludeScrollbar,
-            passiveListener: (config && config.passiveListener === false) ? false : true
+            passiveListener: config && config.passiveListener
           };
         },
 
@@ -227,9 +227,17 @@
             if (!events.forEach) {
               events = [events];
             }
-            var options = isPassiveListenerSupported() ? { passive: this.props.passiveListener } : false;
+            var options;
+            if (typeof this.props.passiveListener === 'boolean' && isPassiveListenerSupported()) {
+              options = { passive: this.props.passiveListener };
+            }
             events.forEach(function (eventName) {
-              document.addEventListener(eventName, fn, options);
+              if (options) {
+                document.addEventListener(eventName, fn, options);
+              }
+              else {
+                document.addEventListener(eventName, fn);
+              }
             });
           }
         },
--- a/index.js
+++ b/index.js
@@ -114,7 +114,8 @@
 
         getDefaultProps: function() {
           return {
-            excludeScrollbar: config && config.excludeScrollbar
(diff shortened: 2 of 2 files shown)


Pomax:
Can you explain what problem this solves? Because I can't think of a real world situation where this makes an actual difference, as outside clicks aren't continuously monitored for, they're only monitored for while an element is on the page that requires outside click listening, which is pretty much mutually exclusive with "scrolling page content".

mrblackus:
I had the problem when trying to detect clicks outside a sidebar panel. When the panel is opened, if I click on the document content, the panel closes.

If the click that is outside the panel is done on a button or a link, the action of this element is performed, which in my case was not what I wanted (not sure the reason of that is relevant here, I can elaborate if you want).

For such a case, I want to be able to stop the event here by using `event.preventDefault()` on the handler. As the fired event was a `touch` one (I was working on a mobile browser), I couldn't call `preventDefault()` ont it because the event listener is passive by default.

This PR give the possibility to make the event listeners not passive, to cover such unusual situations.

Pomax:
I'm still slightly confused, why are you not using the ignore class on those elements? That's what it's for. https://github.com/Pomax/react-onclickoutside#marking-elements-as-skip-over-this-one-during-the-event-loop

mrblackus:
I thought about using it, but as the sidebar (and so the panel when opened) is always displayed on the screen (fixed position), I didn't want to add the class on every interactive elements on the whole website.

I consider it is more clean to avoid the events being propagated rather than adding a specific CSS class (that has no effect on styling) on every elements that could be clicked in this situation. It avoids making all these components "dependent" of this class.

Pomax:
In that case can you please file an issue that explains your situation, so we can tie this PR to that, and then add some test code that makes sure that what you actually set out to achieve is in fact achieved by this code in a way that `npm test` can verify? I still don't understand why what you need to do can only be done cleanly by relying on an experimental new DOM event property, so having PoC code with interaction testing to illustrate the validity of your use case is important.

As concrete PR comment, your code changes events to `passive: true` in the absence of an override, which sounds like a really bad idea: according to the spec this change means `preventDefault()` won't work, and your PR makes that the *default* behaviour, which makes the library backward-incompatible. Instead, this should be "false, unless explicitly set to true".

mrblackus:
Oh indeed, I wrote it a little fast. I should not give an option to the event listener at all if nothing is given as an option. I'll fix that.

That being said, passive event listeners are pretty well supported ([CanIUse](http://caniuse.com/#feat=passive-event-listener)) and are default on Chrome since version 56 which is pretty well adopted according to CanIUse (I don't know how their adoption metrics are reliable).

So I will give nothing to `addEventListener` is no option is given, and *false* if false is explicitly given.
So the browser default will always be used, unless explicitly stated (which is what I wanted). Introducing a breaking changes by changing default behaviour was not my will.

Pomax:
I still want to understand why "not being able to cancel the default browser behaviour" makes sense here, though. By default, the library does not call `evt.preventDefault()` unless you explicitly passed in `preventDefault={true}`, so having passive events shouldn't have any effect unless you have a really weird codebase where you need preventDefault **and** you need it to not happen, which would definitely be a sign of bad code.

mrblackus:
Hello there, sorry for the inactivity.

I think I didn't expose clearly my situation, that's probably why it seems so obscure.

I have a sidebar panel on the left of my screen. When it's opened the onClickOutside handler is active.
When clicking outside the panel, I want it to close, which is working perfectly.

Problem is when I click outside of the panel and that click lands on an interactive element (on a button, for example). On my case, I don't want the button (or anything on the document) to be clicked when clicking outside the panel, so I want to drop the event calling `preventDefault()`.

That's here when the passive event enter in the game. Since my event listener is passive (because it's a `touch` one) by default (Chrome 56), I can't call `preventDefault()` to stop my event right here.

This is why I need to be able to flag the event listener as not passive, so that I can stop its propagation.

I hope my explanation is more clear like that.

Pomax:
Then you don't want (or really: need) this HOC. You want your menu to sit at some z-index "height", and add a plain old 100% width and height transparent underlay for your menu at z-index "menu minus 1", and then users clicking/touching "outside" the menu actually registers on the underlay, not the content under that.

That's also much easier than using this HOC using some code like:

```
class MyMenuWithUnderlay extends React.Component {
  hide() {
    this.setState({ hidden: true });
  }
  render() {
    if (this.state.hidden) return null;
    return <div>
      <Menu options={...} />
      <div className="menu-underlay" onClick={() => this.hide()} />
    </div>
  }
}
```

with some associated CSS:

```
.menu { display: block; z-index: 100; }
.menu-underlay { display: fixed; top: 0; right: 0; bottom: 0; left: 0; z-index: 99; background: transparent; }
```

And we're done.

Pomax:
obviated by https://github.com/Pomax/react-onclickoutside/commit/784b404bc388abc62d7c9aea3e4aba8fc78c5c45

Issue #159, opened 2017-03-21
onClickOutside doesn't respond to Composed component's updated props


121watts:
Hi all!  Thanks for the great package.  We use it all over the place in our app here @ influxdata.  

TLDR;

Should I be handling stateless components differently than in the example I provide?  
Or, could react-onclickoutside be updated to respond to composed component prop changes?

------
I have a stateless component that looks (kinda) like this:

```
const DatabaseRow = ({
  retentionPolicy,
}) => {
const {name, duration, replication, isEditing, isDefault, isNew} = retentionPolicy
  return (
    <div>
      `${name}${duration}${replication}${isEditing}${isDefault}${isNew}`
    </div>
  )
}

export default onClickOutside(DatabaseRow)
```

Which gets rendered by another component which looks a bit like this:

```
database.retentionPolicies.map(retentionPolicy => {
  return (
    <DatabaseRow
      key={rp.id}
      handleClickOutside={() => onStopEditRetentionPolicy({database, retentionPolicy})}
     />
  )
})
```

When I updated `retentionPolicy`, I'd expect `handleClickOutside` to create a new closure around the updated `retentionPolicy`.  So, when `handleClickOutside` fires `onStopEditRetentionPolicy` can operate on the most up to date `retentionPolicy`. 

Here's what I think is happening:

It looks like [this line](https://github.com/Pomax/react-onclickoutside/blob/master/index.js#L148) assigns the closure around `onStopeEditRetentionPolicy` (and its arguments) to the `clickOutsideHandler` variable on `componentDidMount`.  It looks like that variable is not updated / reassigned when the Composed component's props update.  I worked around this by copying all of the code in `componentDidMount` into `componentDidUpdate`.  Should I be handling stateless components differently?  Or, could react-onclickoutside be updated to respond to composed component prop changes?

Pomax:
Hm, isn't the function signature for a stateless component `props => { ...; return ...; }`? It looks like your component actually follows a different signature, with an implicit object assignment that has the same property twice?

121watts:
@Pomax the syntax I used is a bit confusing, but is valid.  Apologies.  I updated it to a more standard syntax and this issue persists.  Thank you for the fast response!

Pomax:
Hm, in the interest of a test, can you see what happens if you use
```
const DatabaseRow = (props) => {
  const {name, duration, replication, isEditing, isDefault, isNew} = props
  return (
    <div>
[...]

PR #169, opened 2017-04-22
Es6 class

Andarist:
Please review if it meets requirements. Ive done the best to extent of my knowledge about those. 

Aint sure how you are going approach this now as you wanted to include it for v6 when React16 is released. Maintaining 2 such different branches now will be quite a pain. Maybe we could release this as `next` from `master` and keep recent `master` as `5.x` branch and keep it as default for npm. It would require cherry-picking back and forth for some time, but there is not that much of active development going here, so maybe it wont be that bad after all. WDYT?

The only thing Im not so sure about is that I got rid of `findDOMNode` in favor of callback ref. But it might not cover 100% of use cases

Also didnt yet fixed the test task

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "plugins": ["transform-class-properties"]
+}
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,0 @@
-{
-  "plugins": ["transform-class-properties"]
-}
--- a/package.json
+++ b/package.json
@@ -61,6 +61,10 @@
     "rollup-plugin-node-resolve": "^3.0.0",
     "webpack": "^1.15.0"
   },
+  "peerDependencies": {
+    "react": "^15.5.x",
+    "react-dom": "^15.5.x"
+  },
   "babel": {
     "plugins": [
       "transform-class-properties"
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
   ],
   "homepage": "https://github.com/Pomax/react-onclickoutside",
   "authors": [
-    "Pomax <pomax@nihongoresources.com>"
+    "Pomax <pomax@nihongoresources.com>",
+    "Andarist <mateuszburzynski@gmail.com>"
   ],
   "keywords": [
     "react",
@@ -26,9 +27,11 @@
   },
   "scripts": {
     "build": "rollup -c -i src/index.js -o ./index.js",
-    "test": "npm run lint && npm run build && karma start test/karma.conf.js --single-run && npm run test:nodom",
-    "test:nodom": "mocha test/no-dom-test.js",
     "lint": "eslint src/*.js ./test",
+    "test": "run-s test:**",
+    "test:basic": "run-s lint build",
+    "test:karma": "karma start test/karma.conf.js --single-run",
+    "test:nodom": "mocha test/no-dom-test.js",
     "precommit": "npm run test"
   },
   "devDependencies": {
@@ -46,6 +49,7 @@
     "karma-spec-reporter": "0.0.26",
     "karma-webpack": "^2.0.2",
     "mocha": "^3.2.0",
+    "npm-run-all": "^4.0.2",
     "phantomjs-prebuilt": "^2.1.7",
     "react": "^15.5.x",
     "react-dom": "^15.5.x",
@@ -56,5 +60,10 @@
     "rollup-plugin-babel": "^2.7.1",
     "rollup-plugin-node-resolve": "^3.0.0",
     "webpack": "^1.15.0"
+  },
+  "babel": {
+    "plugins": [
+      "transform-class-properties"
+    ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "react-onclickoutside",
   "version": "5.11.1",
   "description": "An onClickOutside wrapper for React components",
-  "main": "index.js",
+  "main": "./index.js",
   "files": [
     "index.js"
   ],
@@ -25,20 +25,21 @@
     "url": "https://github.com/Pomax/react-onclickoutside/issues"
   },
   "scripts": {
-    "test": "npm run lint && karma start test/karma.conf.js --single-run && npm run test:nodom",
+    "build": "rollup -c -i src/index.js -o ./index.js",
+    "test": "npm run lint && npm run build && karma start test/karma.conf.js --single-run && npm run test:nodom",
     "test:nodom": "mocha test/no-dom-test.js",
-    "lint": "eslint index.js ./test",
-    "precommit": "npm run lint"
-  },
-  "dependencies": {
-    "create-react-class": "^15.5.x"
+    "lint": "eslint src/*.js ./test",
+    "precommit": "npm run test"
   },
   "devDependencies": {
-    "babel-preset-es2015": "^6.14.0",
+    "babel-cli": "^6.24.1",
+    "babel-eslint": "^7.2.3",
+    "babel-loader": "^6.x",
+    "babel-plugin-transform-class-properties": "^6.24.1",
+    "babel-preset-es2015": "^6.24.1",
     "chai": "^3.5.0",
     "eslint": "^3.4.0",
     "karma": "^1.4.0",
-    "karma-babel-preprocessor": "^6.0.1",
     "karma-chai": "^0.1.0",
     "karma-mocha": "^1.3.0",
     "karma-phantomjs-launcher": "^1.0.0",
@@ -47,10 +48,13 @@
     "mocha": "^3.2.0",
     "phantomjs-prebuilt": "^2.1.7",
     "react": "^15.5.x",
-    "react-addons-test-utils": "^15.5.x",
     "react-dom": "^15.5.x",
     "react-test-renderer": "^15.5.x",
     "require-hijack": "^1.2.1",
-    "webpack": "^1.12.14"
+    "rimraf": "^2.6.1",
+    "rollup": "^0.41.6",
+    "rollup-plugin-babel": "^2.7.1",
+    "rollup-plugin-node-resolve": "^3.0.0",
+    "webpack": "^1.15.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "test:basic": "run-s lint build",
     "test:karma": "karma start test/karma.conf.js --single-run",
     "test:nodom": "mocha test/no-dom-test.js",
-    "precommit": "npm run test"
+    "postinstall": "node test/setup-pre-commit-hook"
   },
   "devDependencies": {
     "babel-cli": "^6.24.1",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,0 +1,22 @@
+import nodeResolve from 'rollup-plugin-node-resolve';
+import babel from 'rollup-plugin-babel';
+
+var config = {
+  format: 'umd',
+  moduleName: 'onClickOutside',
+  exports: 'named',
+  external: ['react'],
+  globals: {
+    react: 'React'
+  },
+  plugins: [
+    nodeResolve({
+      jsnext: true
+    }),
+    babel({
+      exclude: 'node_modules/**',
+    })
+  ]
+};
+
+export default config;
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -5,9 +5,10 @@ var config = {
   format: 'umd',
   moduleName: 'onClickOutside',
   exports: 'named',
-  external: ['react'],
+  external: ['react', 'react-dom'],
   globals: {
-    react: 'React'
+    react: 'React',
+    'react-dom': 'ReactDOM'
   },
   plugins: [
     nodeResolve({
--- a/src/index.js
+++ b/src/index.js
@@ -1,0 +1,197 @@
+import { createElement, Component } from 'react';
+import generateOutsideCheck from './generateOutsideCheck';
+
+/**
+ * A higher-order-component for handling onClickOutside for React components.
+ */
+const registeredComponents = [];
+const handlers = [];
+
+/**
+ * This function generates the HOC function that you'll use
+ * in order to impart onOutsideClick listening to an
+ * arbitrary component. It gets called at the end of the
+ * bootstrapping code to yield an instance of the
+ * onClickOutsideHOC function defined inside setupHOC().
+ */
+export default function onClickOutsideHOC(WrappedComponent, config) {
+  return class onClickOutside extends Component {
+    static displayName = `OnClickOutside(${ WrappedComponent.displayName || WrappedComponent.name || 'Component' })`
+
+    static defaultProps = {
+      eventTypes: ['mousedown', 'touchstart'],
+      excludeScrollbar: (config && config.excludeScrollbar) || false,
+      outsideClickIgnoreClass: 'ignore-react-onclickoutside',
+      preventDefault: false,
+      stopPropagation: false,
+    }
+
+    static getClass = () => WrappedComponent.getClass ? WrappedComponent.getClass() : WrappedComponent
+
+    /**
+     * Access the WrappedComponent's instance.
+     */
+    getInstance() {
+      if (!WrappedComponent.prototype.isReactComponent) {
+        return this;
+      }
+      const ref = this.instanceRef;
+      return ref.getInstance ? ref.getInstance() : ref;
+    }
+
+    // this is given meaning in componentDidMount
+    __outsideClickHandler() {}
+
+    /**
+     * Add click listeners to the current document,
+     * linked to this component's state.
+     */
+    componentDidMount() {
+      // If we are in an environment without a DOM such
+      // as shallow rendering or snapshots then we exit
+      // early to prevent any unhandled errors being thrown.
+      if (typeof document === 'undefined' || !document.createElement){
+        return;
+      }
+
+      const instance = this.getInstance();
+      var clickOutsideHandler;
+
+      if(config && typeof config.handleClickOutside === 'function') {
+        clickOutsideHandler = config.handleClickOutside(instance);
+        if(typeof clickOutsideHandler !== 'function') {
+          throw new Error('WrappedComponent lacks a function for processing outside click events specified by the handleClickOutside config option.');
+        }
+      } else if(typeof instance.handleClickOutside === 'function') {
+        if (Component.prototype.isPrototypeOf(instance)) {
+          clickOutsideHandler = instance.handleClickOutside.bind(instance);
+        } else {
+          clickOutsideHandler = instance.handleClickOutside;
+        }
+      } else if(typeof instance.props.handleClickOutside === 'function') {
+        clickOutsideHandler = instance.props.handleClickOutside;
+      } else {
+        throw new Error('WrappedComponent lacks a handleClickOutside(event) function for processing outside click events.');
+      }
+
+      // TODO: try to get rid of this, could be done with function ref, might be problematic for SFC though, they do not expose refs
+      const componentNode = this.instanceRef;
+      if (componentNode === null) {
+        console.warn('Antipattern warning: there was no DOM node associated with the component that is being wrapped by outsideClick.');
+        console.warn([
+          'This is typically caused by having a component that starts life with a render function that',
+          'returns `null` (due to a state or props value), so that the component \'exist\' in the React',
+          'chain of components, but not in the DOM.\n\nInstead, you need to refactor your code so that the',
+          'decision of whether or not to show your component is handled by the parent, in their render()',
+          'function.\n\nIn code, rather than:\n\n  A{render(){return check? <.../> : null;}\n  B{render(){<A check=... />}\n\nmake sure that you',
+          'use:\n\n  A{render(){return <.../>}\n  B{render(){return <...>{ check ? <A/> : null }<...>}}\n\nThat is:',
+          'the parent is always responsible for deciding whether or not to render any of its children.',
+          'It is not the child\'s responsibility to decide whether a render instruction from above should',
+          'get ignored or not by returning `null`.\n\nWhen any component gets its render() function called,',
+          'that is the signal that it should be rendering its part of the UI. It may in turn decide not to',
+          'render all of *its* children, but it should never return `null` for itself. It is not responsible',
+          'for that decision.'
+        ].join(' '));
+      }
+
+      const fn = this.__outsideClickHandler = generateOutsideCheck(
+        componentNode,
+        clickOutsideHandler,
+        this.props.outsideClickIgnoreClass,
+        this.props.excludeScrollbar,
+        this.props.preventDefault,
+        this.props.stopPropagation
+      );
+
+      const pos = registeredComponents.length;
+      registeredComponents.push(this);
+      handlers[pos] = fn;
+
+      // If there is a truthy disableOnClickOutside property for this
+      // component, don't immediately start listening for outside events.
+      if (!this.props.disableOnClickOutside) {
+        this.enableOnClickOutside();
+      }
+    }
+
+    /**
+    * Track for disableOnClickOutside props changes and enable/disable click outside
+    */
+    componentWillReceiveProps(nextProps) {
+      if (this.props.disableOnClickOutside && !nextProps.disableOnClickOutside) {
+        this.enableOnClickOutside();
+      } else if (!this.props.disableOnClickOutside && nextProps.disableOnClickOutside) {
+        this.disableOnClickOutside();
+      }
+    }
+
+    /**
+     * Remove all document's event listeners for this component
+     */
+    componentWillUnmount() {
+      this.disableOnClickOutside();
+      this.__outsideClickHandler = false;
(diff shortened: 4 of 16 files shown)


Pomax:
I have no real intention of maintaining two different branches, as per the README's point that only the latest version gets maintained.

One thing that surprises me is that your rewrite made the library bigger, rather than smaller. ES6 has a far terser syntax and typically makes codebases shrink, not grow. To that effect, if we want to pursue this now, rather than half a year from now, can I ask you to use rollup? There is no JSX in these files, and modern browsers do not need the code converted to ES5 at all, so just rollup to turn the class collection into a single ES6 bundle without `import` statements will be a much better solution for generating the `dist` library.

.gitignore
```diff
@@ -1 +1,3 @@
 node_modules
+es
+lib
```
Pomax: Transient dirs should not be in the gitignore, they should be removed once the task that needed them finishes. NPM offers the super handy `pre...` and `post...` tasks for this, so if a dir `bar` is used by task `foo`, you can write `"postfoo": "shx rm -rf bar"` and cleanup happens automatically when you `npm run foo`.

src/index.js
```diff
@@ -0,0 +1,195 @@
...
+      }
+    }
+
+    getRef = ref => this.instanceRef = ref
+
+    /**
+     * Pass-through render
+     */
+    render() {
+      const {
```
Pomax: This seems a little strange. `const` incurs additional overhead that we never hit, and the code happily modifieds `props` so allocating it as a const basically "lies" about the intent. Let's change this to `var` (not `let`, because these are not block scoped, but function scoped)

.babelrc
```diff
@@ -0,0 +1,13 @@
+{
+  "env": {
+    "cjs": {
+      "presets": ["es2015", "react", "stage-2"],
```
Pomax: there is no JSX code, and aside from `import` modern browsers, as well as the current LTS of Node itself supports ES6, so I see no reason to do all these things. Just rollup, set to rollup to UMD, should be enough. 

See https://github.com/rollup/rollup/wiki/JavaScript-API but then with the obvious `format: 'umd'`

package.json
```diff
@@ -2,7 +2,9 @@
   "name": "react-onclickoutside",
   "version": "5.11.1",
   "description": "An onClickOutside wrapper for React components",
-  "main": "index.js",
+  "main": "./lib/index.js",
```
Pomax: rather than this, let's have a `src` dir that contains the code, and have it rollup to `./index.js`, then keep everything the same: as a umd module everything should just load that index.js

package.json
```diff
@@ -25,17 +27,25 @@
     "url": "https://github.com/Pomax/react-onclickoutside/issues"
   },
   "scripts": {
+    "build:es": "rimraf es && cross-env BABEL_ENV=es babel src --out-dir es",
+    "build:cjs": "rimraf lib && cross-env BABEL_ENV=cjs babel src --out-dir lib",
+    "build:umd:dev": "cross-env BABEL_ENV=es NODE_ENV=development rollup -c -i src/index.js -o dist/react-onclickoutside.js",
+    "build:umd:prod": "cross-env BABEL_ENV=es NODE_ENV=production rollup -c -i src/index.js -o dist/react-onclickoutside.min.js",
+    "build:umd": "rimraf dist && npm run build:umd:dev && npm run build:umd:prod",
+    "build": "npm run build:es && npm run build:cjs && npm run build:umd",
```
Pomax: let's not generate all these separate files. Just rollup the source to a umd bundle and we're done.

package.json
```diff
@@ -25,17 +27,25 @@
...
     "test": "npm run lint && karma start test/karma.conf.js --single-run && npm run test:nodom",
     "test:nodom": "mocha test/no-dom-test.js",
     "lint": "eslint index.js ./test",
-    "precommit": "npm run lint"
-  },
-  "dependencies": {
-    "create-react-class": "^15.5.x"
+    "precommit": "npm run lint && npm run build"
```
Pomax: this should run `npm test`

rollup.config.js
```diff
@@ -0,0 +1,45 @@
...
+    }),
+    replace({
+      'process.env.NODE_ENV': JSON.stringify(env)
+    })
+  ]
+}
+
+if (env === 'production') {
+  config.plugins.push(
+    uglify({
```
Pomax: this is not necessary, and in fact counter-productive: as a library that is itself used in larger bundles, *those* bundles are responsible for the minification, and can do a far better job at that when this library is not pre-uglified with its own settings.

.gitignore
Andarist: but those dirs are needed for `npm publish` i guess, however are created by a `build` , its easy to run a build just to check if everything is alright without need to publish

src/index.js
Andarist: i understand ur comment about 'lie' and gonna change it, aint sure however why this should be `var` here, as this is declared at the top of the function `let` and `var` would behave basically the same here, and using `var` nowadays means to the reader that there is something special going here, which is not

.babelrc
Andarist: ive used `return <WrappedComponent {...props} />;` in the `render` function, which ofc could be avoided if u wish

package.json
Andarist: then every user of `webpack2` or `rollup` would have to include in their apps the `umd` bloat (require wrapper) for no good reason, this allows every bundler to chose the best (minimal) build

rollup.config.js
Andarist: ye sure, but thats what `es` and `lib` dirs are there for, this is just creating `umd` build for the `dist` folder so it can be used just as script which I thought is a requirement, because the way lib was constructed before

Andarist:
@Pomax 
> I have no real intention of maintaining two different branches, as per the README's point that only the latest version gets maintained. So we either do this work once 16 is about to come out, or we replace the master branch so that index.js is always generated as part of npm test based on the ES6 code.

replacing the `master` would be the preferable way from my point of view, less work in the future

> One thing that surprises me is that your rewrite made the library bigger, rather than smaller. ES6 has a far terser syntax and typically makes codebases shrink, not grow. To that effect, if we want to pursue this now, rather than half a year from now, can I ask you to use rollup? There is no JSX in these files, and modern browsers do not need the code converted to ES5 at all, so just rollup to turn the class collection into a single ES6 bundle without import statements is enough. No need for babel or webpack here (webpack is only used in the test suite, and there should be no reason to pull it in for the regular code path)

Well, by including `class` transforms alone it has produced some extra code and how should I approach this without `babel`? The only thing that comes to my mind which could be done is to use `loose` mode so it produces less code.

.gitignore
Pomax: They are not needed if there is only one file that already gets built because it's needed for testing, and so can be checked in just fine.

src/index.js
Pomax: Not really. If you know JS, and you've learned the difference between `var`, `let`, and `const`, then you know that `var` is the function scoped, which is what we're declaring here, if you see `let`, you are looking at a block scoped variable, and if you only see code that uses `let`, you know you're looking at a highly suspect codebase. Just like if the code only uses templating strings.

.babelrc
Pomax: please, yes. Adding in JSX means we suddenly need many more MBs in dependencies and precompile conversion for a single line of code =)

package.json
Pomax: Always test your convictions against the real world: the "bloat" you talk about is 263 bytes. That is *not* going to impact anyone, while allowing this library to ditch 10+ dependencies that this PR adds for creating the separate files.

Pomax:
> Well, by including class transforms alone it has produced some extra code and how should I approach this without babel?

Simple: *you* don't, or really, *we* don't.

Modern browsers and Node.js have supported classes for literally years now (most browsers have supported them since 2014, even Edge has supported them since mid-2015) so the only problem with ES6 classes is the `import` statement, which rollup solves. So we use that, we don't bother with any transforms for legacy systems, we mark it as V6 and we update the README.md to say "if you need ES5 support, either use 5.x or make sure your bundle is ES5ified".

rollup.config.js
Pomax: as stated previously, the intention is to still only generating one, umd library.

Andarist:
So I've changed few things. Got tests to work, had to make few adjustments, where possibly some could be avoided, but couldnt make it work with previous settings.

> Always test your convictions against the real world: the "bloat" you talk about is 263 bytes. That is not going to impact anyone, not even if they relied on 25 rollup umb libraries.

This is not true, at least in my eyes ofc. It would take 6kb~ which is not nothing and from my perspective it's always better to include less than more so less code can be shipped over the wire to the user.

> On the other hand, it allows this library to ditch 10+ dependencies that this PR adds for creating the separate files, as well as remove the need multiple files that are all "the same library", so that is a perfectly acceptable tradeoff.

Is dev dependency really a cost that should guide in such manner? I always see them as free if its all automated and helps devs to ship better code or less code and some extra megabytes on dev machine is acceptable tradeoff in my opinion.

> the assumption that the library is used in larger bundles that are responsible for shimming, uglification, optimization, etc. There is zero reason to do that prematurely here.

If that's the ur point of view I dont even see why this should be bundled at all as UMD, if its supposed to be taken care of by a bundler in somebody's application let them chose how they want to bundle it. Obvious downside is ofc that people have various babel settings and that's certainly not my opinion on the matter, but I'm feeling like such statement is a little bit inconsistent with a desire to have this UMDified.

> Modern browsers and Node.js have supported classes for literally years now (most browsers have supported them since 2014, even Edge has supported them since mid-2015) so the only problem with ES6 classes is the import statement, which rollup solves. So we use that, we don't bother with any transforms for legacy systems, it gets released as v6.x and we update the README.md to say "if you need ES5 support, either use 5.x or make sure your bundle is ES5ified" in the same way it already tells people which versions to use based on which React they're using.

If you have luxury of being able to ship es6 code to the browsers I really envy you, but most of applications still need to transpile this as buisness requirement. I find myself fixing bugs for ie10/11 and mobile safari quite often. Burdening users to add some extra rules so this lib can be ES5ified is quite cumbersome I think.

For me that's not the problem as I'm involved now in pushing this forward and I'm well informed, but let's face it - people dont read too much of the README before starting to use a library and when people are developing their code rather using modern browsers they can not notice that it might break on legacy browsers before shipping it to the user.

If your opinion is that many optimizations happen in the userland anyway this extra helpers for creating classes etc is also not a cost as it will get deduped by some webpack plugin anyway.

Out of curiosity Ive check every dependency I have and literally none of them uses your proposed approach. For me its quite a signal that we are truly not yet in a place when we (as developers) can ship es6 to the browser if we are targeting a global scale.

> as stated previously, the intention is to still only generating one, umd library.

You are the author of this library and ultimately you have the last voice in this matter, so I won't impose my opinions too much, just please take them into a consideration and let me know how should I tweak this PR so it gets merged in.

My proposal is to ditch `dist` files, nearly noone uses them anyway. Skip `lib`, bundle umd to the `index.js`, but keep `es` so webpack2 and rollup can work with `jsnext:main`/`module` format.

Pomax:
>So I've changed few things. Got tests to work, had to make few adjustments, where possibly some could be avoided, but couldnt make it work with previous settings.

Awesome! I'll review this when I get home today, but in the interrim, which settings didn't work?

> This is not true, at least in my eyes ofc. It would take 6kb~ which is not nothing and from my perspective it's always better to include less than more so less code can be shipped over the wire to the user.

263 bytes is an irrelevant increase on the total bundle size in terms of filesystem use, and it gets even less significant when actually used on the web, because any sane server sends data with gzip compression, which makes that 263 *truly* tiny: taking this library as example, the difference between the gzipped version and the gzipped with-umd-header version is three bytes. So let's cap this: using a single UMD library is fine, and v6 of this library will still be a single file with as little building as necessary to yield a single ES6 umd library.

> Is dev dependency really a cost that should guide in such manner? I always see them as free if its all automated and helps devs to ship better code or less code and some extra megabytes on dev machine is acceptable tradeoff in my opinion.

Yes.

> If you have luxury of being able to ship es6 code to the browsers I really envy you,

Why? You have the same luxury. 

In a nutshell, so we can stop having this conversation: updates to this library will only ever be for "the current React at the time of the update, for the current browser landscape at the time of the update". If someone needs support for legacy browsers, they have two options:

1. peg library versions in package.json to known-working-for-legacy, or
2. set up your own build system so that you generate a bundle that is legacy-compatible, either in the same file or as separate bundle (and realistically, any half decent code base already does this).
 
The README.md has a section on version compatibilities, every time a major version comes out that section gets expanded to explain what's what for people who need to use this library.

> Out of curiosity Ive check every dependency I have and literally none of them uses your proposed approach.

I'm okay with that in 2017.

Andarist:
> Awesome! I'll review this when I get home today, but in the interrim, which settings didn't work?

couldnt make it work with `preprocessors` set to `babel` + `webpack`, so i chose to use `webpack` alone and have it load files with `babel`, so I dont get any ordering issues now

> Why? You have the same luxury.

not rly :P non-es6 browsers still holds too much market share for us to ignore it

Anyway - gonna update PR later to bundle this how you like it.

Andarist:
@Pomax ok, i've updated the code, please let me know if everything is alright or should something be corrected

Pomax:
Apoologies for the delay, work had me focussed on very different things for a while.

>not rly :P non-es6 browsers still holds too much market share for us to ignore it 

my point was you, just like me, have the luxury to *write* as much ES6-only code as you like, because the only place a good build chain performs ES5 conversion is "at the very end", to guarantee the *final* bundle works for the intended target.

Doing any kind of conversion before that moment is most assuredly premature optimization, and comes burdened with all the inefficiencies that premature optimization brings with it due to running on a far smaller total code base. 

Of course, if your target is ES5-only browsers then you need that final step, but you don't need it here, you need it in your own project, when you type `npm run build` or whatever your build command is =)

package.json
```diff
@@ -25,20 +25,21 @@
...
+    "build": "rollup -c -i src/index.js -o ./index.js",
     "test": "npm run lint && karma start test/karma.conf.js --single-run && npm run test:nodom",
     "test:nodom": "mocha test/no-dom-test.js",
-    "lint": "eslint index.js ./test",
-    "precommit": "npm run lint"
-  },
-  "dependencies": {
-    "create-react-class": "^15.5.x"
+    "lint": "eslint src/*.js ./test",
+    "precommit": "npm run build && npm run test"
```
Pomax: we want "npm run build" to be part of the "test" script, so that we're always testing against the result of a build, rather than only doing so during precommit.

Andarist:
No worries about the delay. We are in no rush here :) review it in ur free time

src/index.js
```diff
@@ -0,0 +1,194 @@
...
+    componentDidMount() {
+      // If we are in an environment without a DOM such
+      // as shallow rendering or snapshots then we exit
+      // early to prevent any unhandled errors being thrown.
+      if (typeof document === 'undefined' || !document.createElement){
+        return;
+      }
+
+      const instance = this.getInstance();
+      let clickOutsideHandler;
```
Pomax: we're in function scope. This is a `var` not a `let`. We also want to do something reasonable here if it turns out that `instance === this`.

src/index.js
```diff
@@ -0,0 +1,194 @@
...
+          clickOutsideHandler = instance.handleClickOutside;
+        }
+      } else if(typeof instance.props.handleClickOutside === 'function') {
+        clickOutsideHandler = instance.props.handleClickOutside;
+      } else {
+        throw new Error('WrappedComponent lacks a handleClickOutside(event) function for processing outside click events.');
+      }
+
+      // TODO: try to get rid of this, could be done with function ref, might be problematic for SFC though, they do not expose refs
+      const componentNode = this.instanceRef;
```
Pomax: given that we already fetch a `const instance` on L53, I'm not sure we need another const here.

src/index.js
```diff
@@ -0,0 +1,194 @@
...
+      stopPropagation: false,
+    }
+
+    static getClass = () => WrappedComponent.getClass ? WrappedComponent.getClass() : WrappedComponent
+
+    /**
+     * Access the WrappedComponent's instance.
+     */
+    getInstance() {
+      return WrappedComponent.prototype.isReactComponent ? this.instanceRef : this;
```
Pomax: This needs the same kind of fallthrough as getClass, though:
```
getInstance() {
  if (!WrappedComponent.prototype.isReactComponent) {
    return this;
  }
  const ref = this.instanceRef;
  return ref.getInstance ? ref.getInstance() : ref;
}

```

src/index.js
```diff
@@ -0,0 +1,194 @@
...
+      } else if(typeof instance.props.handleClickOutside === 'function') {
+        clickOutsideHandler = instance.props.handleClickOutside;
+      } else {
+        throw new Error('WrappedComponent lacks a handleClickOutside(event) function for processing outside click events.');
+      }
+
+      // TODO: try to get rid of this, could be done with function ref, might be problematic for SFC though, they do not expose refs
+      const componentNode = this.instanceRef;
+      if (componentNode === null) {
+        console.warn('Antipattern warning: there was no DOM node associated with the component that is being wrapped by outsideClick.');
```
Pomax: based on React's warning pattern, we probably want to make this a "warn once" construction, but that can be done in a follow-up issue instead. Leaving as comment so it can be filed as part of landing this PR

src/index.js
```diff
@@ -0,0 +1,194 @@
...
+          'the parent is always responsible for deciding whether or not to render any of its children.',
+          'It is not the child\'s responsibility to decide whether a render instruction from above should',
+          'get ignored or not by returning `null`.\n\nWhen any component gets its render() function called,',
+          'that is the signal that it should be rendering its part of the UI. It may in turn decide not to',
+          'render all of *its* children, but it should never return `null` for itself. It is not responsible',
+          'for that decision.'
+        ].join(' '));
+      }
+
+      const fn = this.__outsideClickHandler = generateOutsideCheck(
```
Pomax: Hm, if we're only using `fn` once, then I don't think we need the `const fn` here at all. We can just assign to `this.__outsideClickHandler` and then on line 105 say `handlers[pos] = this.__outsideClickHandler` without anything changing.

src/index.js
```diff
@@ -0,0 +1,194 @@
...
+    componentWillReceiveProps(nextProps) {
+      if (this.props.disableOnClickOutside && !nextProps.disableOnClickOutside) {
+        this.enableOnClickOutside();
+      } else if (!this.props.disableOnClickOutside && nextProps.disableOnClickOutside) {
+        this.disableOnClickOutside();
+      }
+    }
+
+    /**
+     * Remove the document's event listeners
```
Pomax: Just for correctness, "Remove all document's event listeners for this component"

src/index.js
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * Can be called to explicitly disable event listening
+     * for clicks and touches outside of this element.
+     */
+    disableOnClickOutside = () => {
+      const fn = this.__outsideClickHandler;
+      if (typeof document !== 'undefined') {
+        let events = this.props.eventTypes;
```
Pomax: We probably need to do some additional tracking for eventType un/listening in componentWillReceiveProps too, given what the code's doing here, so that if the eventTypes change, we unlisten for removed types, and start listening for new types.

src/index.js
```diff
@@ -0,0 +1,194 @@
...
+      }
+    }
+
+    getRef = ref => this.instanceRef = ref
+
+    /**
+     * Pass-through render
+     */
+    render() {
+      var props = Object.keys(this.props)
```
Pomax: hmm, we can probably put this rewriting in the willMount and willUpdate functions, so that we cache the result of this filter/reduce and property extension on incoming props as, say, `this.__cached_props` or something and then pass those on during render instead, so we don't rebuild an invariable during render

package.json
Andarist: i will add build to test task then and remove build from precommit

src/index.js
Andarist: isnt on the `master` and ive tried to focus the PR on refactoring to es6 class only, gonna add it though

src/index.js
Andarist: aint sure how you judge what kind of scope do we have for a variable, i understand them all, but just quite dont get how you pick function's one here

whats the danger of `instance === this` here? im not that familiar with all implications of the codebase

src/index.js
Andarist: on the master:
L53: `var instance = this.getInstance();`
L74: `var componentNode = ReactDOM.findDOMNode(instance);`

I got rid of ReactDOM call here, but it might have been the bad idea. Later in the render we have such block:
```javascript
      if (WrappedComponent.prototype.isReactComponent) {
        props.ref = this.getRef;
      } else {
        props.wrappedRef = this.getRef;
      }
```
and now it will break, as if `WrappedComponent` is a class one we will get access to its instance and not to the underlaying DOM node, we could skip the if and just expose `wrappedRef` to both class and functional components, so they can attach ref to whatever node they wish, but thats certainly a breaking change and because of the new API aint sure if we should include this.

Please decide what should be done here, but the most reasonable thing seems to be reverting the change and keeping `ReactDOM.findDOMNode`

src/index.js
Andarist: Fix for this is already waiting for your review - https://github.com/Pomax/react-onclickoutside/pull/170

So I wouldnt really do anything regarding this warning, but just get approval and merge both PRs

src/index.js
Andarist: such code is on the `master` and didnt change it in any way, although it looked a little weird to me, thought it could have been structured this way to save characters later

src/index.js
Andarist: this can be filed as issue after this PR, and I might find some time to tackle this later on

src/index.js
Andarist: this also should be done in separate PR, Im planning to address this issue https://github.com/Pomax/react-onclickoutside/issues/159#issuecomment-296665883 when  this PR gets merged in so we can avoid troublesome merging later. Can combine the fix with the proposed change by using `shallowEqual` checks in both cases and handle both accordingly.

Andarist:
I've ammended the PR with some of the proposed changes, please review it again and answer some of the questions so I know how to proceed further.

> This is looking pretty good! There's a few things that still need a bit of tightening up, and we'll need to add a bit to the README.md as part of this PR so that it doesn't land without the associated documentation already in place.

What info should be added to the README?

Also - I've removed `componentInstance` argument from `generateOutsideCheck` as Ive noticed its not used by it at all.

src/index.js
Andarist: done

Pomax:
Ugh, sorry for the massive delay. The README.md should have two updates:

In the intro:
```
This HOC supports stateless components as of v5.7.0, and uses pure class notation as of v6.
```

and then in https://github.com/Pomax/react-onclickoutside#which-version-do-i-need-for-which-version-of-react:
```
...

If you use React 15, you can use v4.x, which offers both a mixin and HOC, or use v5.x, which is HOC-only.

If you use React 15.5 (or higher), you can use **v5.11.x**, which relies on `createClass` as supplied by `create-react-class` rather than `React.createClass`, or you can use **v6.x**, which is a rewrite of the HOC that uses pure class notation.
```

Andarist:
@Pomax Ok, i'll leave adding those notes to you - you also can add commits directly to my branch if you want and review the whole thing once again here.

Once it lands I will rebase other pending PRs and fix https://github.com/Pomax/react-onclickoutside/issues/159 . Was waiting with this to avoid conflicts.

Pomax:
I've filed https://github.com/Andarist/react-onclickoutside/pull/1 against your PR, but while karma/mocha tests pass I am seeing the manual browser tests 2 and 3 fail, so there might be some work left to figure out what's going wrong there.

Andarist:
ok, we'll try to check out browser tests when

Pomax:
@Andarist it looks like your comment may have got cut off?

Andarist:
should be without 'when', nevermind ;p gonna checkout this today/2morrow

Andarist:
@Pomax should be ok now

Pomax:
@Andarist looks like I'm equally guilty of not running `npm test`. Let me file one more PR against yours with fixes that appease eslint

Pomax:
Filed as https://github.com/Andarist/react-onclickoutside/pull/2

Andarist:
Great! Yeah, ive kinda focused on fixing those browser tests and forgot to run units >.<

Everything is merged into this PR. I think the work is done and we can merge this into the master :) Maybe without releasing this yet - let's tackle other issues with passive listeners etc and then release v6, as releasing it is a good moment for some possible breaking changes.

Tip - you can commit directly on my branch :)

Pomax:
I prefer not to directly commit: it's always better to have that second pair of eyes before you proceed with a merge. Too many accidentally bad direct commits in the past for me to ever write one again. It's PRs all the way down =P

As for not releasing yet: fully agreed. A new release can be spun once we have some of the other oustanding issues merged in, too.

Pomax:
merged - awesome job @Andarist!

ehd:
@Pomax @Andarist Thank you so much!

PR #170, opened 2017-04-24
Fix/161

### LINKED ISSUE #161: React components can return null, "Return null" warning not necessary
The warning that pops up when a component returns a `null` value is not necessary.

See here:  
https://facebook.github.io/react/docs/conditional-rendering.html#preventing-component-from-rendering

Andarist:
Fixes #161


Andarist:
@Pomax rebased after `es6-class` merge

test/test.js
```diff
@@ -292,16 +293,57 @@ describe('onclickoutside hoc', function() {
   });
 
   describe('with child rendering as null', function() {
-    var StatelessComponent = () => null;
```
Pomax: This should still work and be tested for: "not even a React component" stateless generators that are just functions with a single argument, returning JSX (or, createElement equiv of course) are still fully supported by React, so testing for that should *not* be removed.

Rather than rewriting this test, we should add the new code as a new test case in addition to the one that was already here. We might want to rename the tests a little so that testing the null-generator has a reasonable name, but almost any time you remove (even if by rewriting) a test case, you're removing the ability to catch regression

Andarist: Hm, aint sure if I follow what u mean exactly - but I'll restore the test with switched asserts in try/catch.

Pomax: I meant you took a test, and then changed how it worked. That is now how you maintain tests. Instead, copy-paste the test, leave the original alone, rename the copy based on what it's going to test,, and then update its code to do the new thing you're testing for.

test/test.js
```diff
@@ -292,16 +293,67 @@ describe('onclickoutside hoc', function() {
...
-      } catch(e) {
-        assert(true, 'an error was thrown');
+        assert(true, 'component was wrapped despite having no DOM node on mount');
+      } catch (err) {
+        assert(false, 'an error was thrown');
       }
     });
-  });
 
+    var counter;
```
Pomax: this code should really be in a `before` block rather than persisting from one test to the next; tests should be able to run regardless of the order they run in. We can land it this way, but we'll need a follow-up to clean up this code so that tests are self-contained.

test/test.js
Andarist: Writing from mobile so i cant rly check the actual code now, but how would u like me to keep old test? Wasnt it testing an incorrect behaviour (now)? At least from whati remember. Wouldnt it just fail if unchanged?

test/test.js
Andarist: Can fix this once i get home. So let it just hang here for a moment longer

PR #172, opened 2017-05-03
Provide explicit passive setting of the listener based on props.preve…

### LINKED ISSUE #168: Unable to preventDefault on Chrome
I am having a bit of trouble with preventDefault in chrome. When I pass `preventDefault={true}` as prop, I get this warning in the chrome and preventDefault fails.

> index.js:68 Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/5093566007214080

I also fails with the same error when I try `event.preventDefault()` inside `handleClickOutside(event)`

The error occurs on all Chrome versions >=54. It works fine on Firefox.

Andarist:
…ntDefault as desktop/mobile have different defaults (fixes #168)

### DIFF
--- a/index.js
+++ b/index.js
@@ -100,7 +100,10 @@ function onClickOutsideHOC(WrappedComponent, config) {
           if (!events.forEach) {
             events = [events];
           }
-          events.forEach(eventName => document.addEventListener(eventName, fn));
+          events.forEach(eventName => {
+            const handlerOptions = !this.props.preventDefault && ['touchstart', 'touchmove'].indexOf(eventName) !== -1 ? { passive: true } : null;
+            document.addEventListener(eventName, fn, handlerOptions);
+          });
         }
       }, this.disableOnClickOutside = () => {
         const fn = this.__outsideClickHandler;
--- a/src/index.js
+++ b/src/index.js
@@ -151,7 +151,12 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
         if (!events.forEach) {
           events = [events];
         }
-        events.forEach(eventName => document.addEventListener(eventName, fn));
+        events.forEach(eventName => {
+          const handlerOptions = !this.props.preventDefault && ['touchstart', 'touchmove'].indexOf(eventName) !== -1
+            ? { passive: true }
+            : null;
+          document.addEventListener(eventName, fn, handlerOptions);
+        });
       }
     }
 


Pomax:
@Andarist it looks like this is failing tests, did you remember to run `npm test` before pushing?

Andarist:
Oh, I had to forget. Maybe we should add `"precommit": "npm run test"`?

I have debugged this and... PhantomJS is broken :/

Chrome:
```
getEventListeners(document).scroll // undefined

var fn = () => {}
document.addEventListener('scroll', fn, {passive:true})

getEventListeners(document).scroll // [Object]

document.removeEventListener('scroll', fn)

getEventListeners(document).scroll // undefined
```

But PhantomJS compares all arguments while `removeEventListener` (and I pass third object param in this PR) against those used in `addEventListener` call. We could tackle this and store reference to this object so it can be passed again to the `removeEventListener` (I've verified that it works then), but should we really? That seems like PhantomJS' issue. Ideally we could setup headless Chrome for testing, but didnt play with it yet.

index.js
```diff
@@ -223,11 +223,12 @@
           var fn = this.__outsideClickHandler;
           if (typeof document !== 'undefined') {
             var events = this.props.eventTypes || DEFAULT_EVENTS;
+            var passive = !this.props.preventDefault;
```
Pomax: This does not feel right. Passive events and preventing browser defaults are different things, even if they work toghter, so entangling them like this is not a good idea.

Andarist: hm, i see `pasive` strongly tied to the `preventDefault` and it seemed reasonable to deduce this here instead of adding yet another aceepted prop

This is not caused by a `click` event but by a `touchstart` which we use here by default - https://www.chromestatus.com/features/5093566007214080 . I guess Chrome decided to treat those as passive by default as `preventDefault` on a touch event prevents scrolling. 

> clicks should absolutely be able to prevent default browser behaviour

That is still possible, just need do this by using `preventDefault` prop instead of calling it manually. OTOH - those seems a little bit iffy to me, what's the reasoning of having those props like `preventDefault` and `stopPropagation` instead of leting the user to handle this on his own? Removing them + adding a `passive` prop would simplify the codebase I think.

Pomax: mostly the fact that events are not cleanly passable things. In any compositional chain there is no guarantee that code won't have done something to the event before the user ever gets their hands on them, so making them *think* they're in control when they aren't is not setting up good user expectation, and I'd like to avoid it.

I think the real solution I'd like to see here is one that adds an exception specifically *only* for touch events. If those are the root cause of the problem, then using "not-preventDefault means it *must* be passive" logic is acceptable, but certainly not for every event, most of which won't ever be passive because they have *nothing* to do with page scrolls (the reason why the passive event concept was invented)

Andarist: Cant really think of a scenario you have mentioned. If somebody is using composition he should be quite aware of what exactly his compositions are doing and when using it on some shared mutable state (like event) manipulation of the passed state is always possible and should be easily recognizable I think.

Maybe just never had any problem with it. Anyway - gonna update a PR in 2 minutes

Andarist:
@Pomax rebased after `es6-class` merge

Pomax:
thanks, @Andarist

PR #176, opened 2017-05-08
Use callback instead of string ref

ehd:
Makes it work with preact and future react versions that might not support string refs.


Pomax:
this has been done in https://github.com/Pomax/react-onclickoutside/pull/169, which will be landing soon.

ehd:
Thanks! I hope it lands soon so I can use preach in production which
doesn't do string refs

PR #178, opened 2017-05-20
Configured prettier for the project

Andarist:
I know you dont like too much dependencies, but thought that those might actually help to improve overall dev experience.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "test:basic": "run-s lint build",
     "test:karma": "karma start test/karma.conf.js --single-run",
     "test:nodom": "mocha test/no-dom-test.js",
-    "postinstall": "node test/setup-pre-commit-hook"
+    "precommit": "npm test && lint-staged"
   },
   "devDependencies": {
     "babel-cli": "^6.24.1",
@@ -42,15 +42,18 @@
     "babel-preset-es2015": "^6.24.1",
     "chai": "^3.5.0",
     "eslint": "^3.4.0",
+    "husky": "^0.13.3",
     "karma": "^1.4.0",
     "karma-chai": "^0.1.0",
     "karma-mocha": "^1.3.0",
     "karma-phantomjs-launcher": "^1.0.0",
     "karma-spec-reporter": "0.0.26",
     "karma-webpack": "^2.0.2",
+    "lint-staged": "^3.4.2",
     "mocha": "^3.2.0",
     "npm-run-all": "^4.0.2",
     "phantomjs-prebuilt": "^2.1.7",
+    "prettier": "^1.3.1",
     "react": "^15.5.x",
     "react-dom": "^15.5.x",
     "react-test-renderer": "^15.5.x",
@@ -68,6 +71,13 @@
   "babel": {
     "plugins": [
       "transform-class-properties"
+    ]
+  },
+  "lint-staged": {
+    "{src,test}/**/*.js": [
+      "prettier --print-width=120 --single-quote --trailing-comma=all --write",
+      "eslint --fix",
+      "git add"
     ]
   }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -17,7 +17,7 @@ const handlers = [];
  */
 export default function onClickOutsideHOC(WrappedComponent, config) {
   return class onClickOutside extends Component {
-    static displayName = `OnClickOutside(${ WrappedComponent.displayName || WrappedComponent.name || 'Component' })`
+    static displayName = `OnClickOutside(${WrappedComponent.displayName || WrappedComponent.name || 'Component'})`;
 
     static defaultProps = {
       eventTypes: ['mousedown', 'touchstart'],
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -69,5 +72,12 @@
     "plugins": [
       "transform-class-properties"
     ]
+  },
+  "lint-staged": {
+    "src/**/*.js": [
+      "prettier --print-width=120 --single-quote --trailing-comma=all --write",
```
Pomax: I like prettier, but is there a way to take out "print-width"? Modern tooling should not care about line lengths. And if you are a vim or emacs user, you are such an incredible power user that you already know how to make those present virtual line breaks without actually saving them into the actual files.

src/index.js
```diff
@@ -51,58 +51,66 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
...
-      const fn = this.__outsideClickHandler = generateOutsideCheck(
+      const fn = (this.__outsideClickHandler = generateOutsideCheck(
         componentNode,
         clickOutsideHandler,
         this.props.outsideClickIgnoreClass,
         this.props.excludeScrollbar,
         this.props.preventDefault,
-        this.props.stopPropagation
-      );
+        this.props.stopPropagation,
```
Pomax: I consider this moronic, but I also don't care.

package.json
Andarist: I think we can only specify the actual width here, cannot opt-out of this - except ofc by passing rly big int or maybe even Infinity? But i aint sure if that works.

Point of prettier is to bring consistency and not every1 is using vim/emacs or have their other editor configured like this

src/index.js
Andarist: Oh, this obe we can skip if u like by seting dangling comma to false, it was just my preference i put here

Pomax: I think it solves the problem of "moving arguments/content around in a container" (whether that's arrays, objects, execution parentheses, etc) at the wrong level. If that is a big problem for you, your probably need to set up your editor such that it isn't a problem anymore.

However, I also don't care: if prettier can enforce this, rather than just lint for it, then let's make it do that.

package.json
Pomax: let's stick in a huge int then. It's 2017, if you are the kind of person who has mastered vim or emacs, you chose to be in control. I see no reason to accomodate limitations in those editors that removed with a few trivial plugin installs.

src/index.js
Andarist: its more a git diff issue rly, without dangling comma whenever u add argument (or move them like u described) u get a little bit more visual bloat while comparing changes (because of comma additions/deletions) than its neccessary

Andarist:
Ive rebased and ammended, so the `print-length` is set to 1024 now, but this one now also reformats chains and smth like this https://github.com/Pomax/react-onclickoutside/pull/178/files#diff-1fdf421c05c1140f6d71444ea2b27638R151 into 1 line. OTOH one could say that this particular line could be refactored to avoid lengthy argument list anyway

Pomax:
hmm. if the choice is between "artificially short lines despite widescreen monitors and it being 2017" and "prettier rewriting code to put chains on a single line", the latter is **by far** worse, so then let's go with a line length of 120. I did not realise this was the tradeoff.

Andarist:
Amended with the 120 width.

package.json
```diff
@@ -69,5 +72,12 @@
     "plugins": [
       "transform-class-properties"
     ]
+  },
+  "lint-staged": {
+    "src/**/*.js": [
```
Pomax: we might as well use `{src,test}/**/*.js` here

package.json
```diff
@@ -42,15 +42,18 @@
     "babel-preset-es2015": "^6.24.1",
     "chai": "^3.5.0",
     "eslint": "^3.4.0",
+    "husky": "^0.13.3",
```
Pomax: we can take this out, it gets installed automatically with lint-staged.

Andarist: hm, according to [this](https://github.com/okonet/lint-staged/blob/c7283b77c2cf3637e9f06f953b919a556ffde205/package.json#L47-L55) it shouldnt, also ive checked `npm install` without this entry and it didnt got installed automatically

package.json
Andarist: done

Andarist:
@Pomax as to releasing v6 I still have like 1 issue to fix, so lets stall this one for a little longer

package.json
Pomax: Hmmmmm.... what do we use it for, then? I don't see any file in `src` or `test` use anything called `husky`, so if our code doesn't use it, and none of the dependencies introduced here use it... why is it a dependency?

Andarist: Thats for automatically running git hooks defined in package.json so its not used anywhere explicitly, but rather it runs itself implicitly

Pomax:
What's the issue you want to fix?

Andarist:
https://github.com/Pomax/react-onclickoutside/issues/159

Pomax:
Unless you already have the PR ready, that should not hold up a v6 release. The current master already solves problems for people on React 15.5, so it's better to release early and release often, than it is to wait until "all the things are in".

I'll push a 6.0.0 and when a PR to address #159 lands we can simply release 6.1.0

PR #180, opened 2017-05-23
Refactored test - counter gets reset in beforeEach hook

Andarist:
Still aint sure how to approach this one - https://github.com/Pomax/react-onclickoutside/pull/170#discussion_r118043966


Pomax:
that's fine, we can iterate on it.

PR #182, opened 2017-05-24
Fix IE not supporting .classList on SVG elements

Andarist:


Pomax:
No.

Andarist:
@Pomax the linked explanation only desribes situation for IE9 and older, however the problem solved by this PR occurs also on IE11 - only for SVG elements. So if I click on SVG icon the lib code automatically throws as .contains access is not guarded.

Pomax:
The first paragraph literally reads "This is true, but also an edge-case problem that only exists for older versions of IE (including IE11), and should be addressed by you, rather than by thousands of individual libraries that assume browsers have proper HTML API implementations (IE Edge has proper classList support even for SVG)."

Andarist:
Oh, sorry - my mobile didnt scroll my down to the linked ID and I got stuck at the top of the README where u mention only the IE9.

Shims are ok if you can pollute global space, but that's not always the case.

While we are before releasing new version of the lib - would u consider dropping default `outsideClickIgnoreClass` and merging in bypassing code before [this if](https://github.com/Pomax/react-onclickoutside/blob/370fefa64d862eef52c2a548a18abe8c1859f714/src/generateOutsideCheck.js#L14-L17) by checking if `ignoreClass` is defined?

Pomax:
I already released 6, so if that commit is part of a PR I'll be happy to review the PR but pending improvements should not hold up releases. Release early, and release often. 

> Shims are ok if you can pollute global space, but that's not always the case.

I don't care about "always", I only care about specific cases. In this particular one, we're talking about a DOM specification issue with is a single, well-defined place where a shim would "fix" things. It does not "polluting the global space", it adds support for a property on a DOM-specific element prototype.

Andarist:
Dropping default `outsideClickIgnoreClass` would be a breaking change though now.

PR #183, opened 2017-05-24
Dropped default outsideClickIgnoreClass, making this now opt-in

Andarist:
As discussed in other PR - this is unfortunately a breaking change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,10 @@ function isNodeFound(current, componentNode, ignoreClass) {
   if (current === componentNode) {
     return true;
   }
+
+  if (typeof ignoreClass === 'undefined') {
+    return false;
+  }
   // SVG <use/> elements do not technically reside in the rendered DOM, so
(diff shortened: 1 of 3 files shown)


Pomax:
Okay, so: why is this change necessary? Is there something broken that this PR fixes, or is it simply a convention change that has the same effect as before? (in which case I'd rather keep things as they were)

Andarist:
Nothing neccessary - just a short circuit for next check which is rather a more specialized case (ignoring certain elements by class name) than a general one.

Which would also out of the box prevent ppl from bumping into IE~ issue. I've proposed it [here](https://github.com/Pomax/react-onclickoutside/pull/182#issuecomment-303840274) shortly after/before v6 release, wanted to make a quick PR out of it. Feel free to discard it if u feel its right thing to do.

Pomax:
Yeah I think I'll close this for now, maybe something to revisit for a future version.

PR #189, opened 2017-05-26
Simplified outside clicks detection by using node#contains

Andarist:


Pomax:
I'm not super happy about this change - if you look at what the code did, it traversed the DOM once, checking parents and classes in a single pass. Your rewrite actually makes this more complex, running a `contains` while *still* traversing the DOM because of the class exception test. 

In this case, while there is a separate `contains` function, it only makes to use it if we can remove the DOM walk entirely. If we can't, running the DOM walk plus a contains check is actually more work than direct parentNode evaluation + class check during the DOM walk.

Andarist:
Yeah, while I was rewriting I thought about this, whole `ignoreClass` prop is screwing with the simplicity of the check. Querying DOM with `querySelector` is not an option either as it brings more complexity too.

We could revisit the idea of `ignoreClass` being an opt-in and follow 2 different paths based on it - simple `.contains()` for default usage and full old traversal for checking both contains (by ===) and classList.

No idea how to simplify it more for now - DOM APIs focus on querying the children rather than 'parents'

Andarist:
@Pomax was thinking about this and personally I would drop `ignoreClass` feature, add smth like `nodesComparator` prop/config option defaulting to `node#contains` + provide in the docs example of its usage for `ignoreClass` use case

Pros:
- simplified containment logic
- being more customizable (any comparator function might get passed in)

Cons:
- breaking backwards compatibility

Pomax:
I think for now the fact that this does not actually simplify the code path means it's better to close this PR until a more efficient solution can be found. If this needs more discussion, let's discuss in an issue, not a PR.

PR #190, opened 2017-05-26
Added prepare script, which can be used by npm5

Andarist:
this can be used when the package is used as git dependency

In theory now there shouldnt be any problem with this, as `build` is hooked as part of the `precommit`, but anyway I think that's useful if somebody forks and changes something or if `build` stops being part of the `precommit` at some point of the time in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     "test:basic": "run-s lint build",
     "test:karma": "karma start test/karma.conf.js --single-run",
     "test:nodom": "mocha test/no-dom-test.js",
-    "precommit": "npm test && lint-staged"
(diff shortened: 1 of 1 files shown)


Pomax:
can you also file an issue so that there's something to tie this PR to, with some text to explain why this is going to be a good idea?

Andarist:
Sure, gonna do that later

Pomax:
I'm a bit unclear - does this PR now close #192, or is #192 a far bigger issue that this PR does not address as a "one issue : one PR" relation?

Andarist:
Sorry for not including this in the explanation - this PR ofc 'fixes'/helps with the issue described in #192

Issue #185, opened 2017-05-25
Need transpiled JS or ES6 src code


Dante-101:
The current v6 is only distributed with ES6 syntax containing `let`, `const`, `class` keywords which breaks the app in old browsers with the error `Uncaught SyntaxError: Use of const in strict mode.`

Babel + Webpack do not transpile the distributed JS of this package in my app but do it for other packages like `react-redux`. I believe this is because `react-redux` distributes transpiled ES5 and ES6 code in different directories and refers the ES6 code using `module: "es/index.js"` and `jsnext:main: "es/index.js"` in `package.json`. See https://github.com/reactjs/react-redux/blob/master/package.json.

Pomax:
I've pushed a v6.0.1 so that there is a "jsnext:main" entry, let me know if that appeases your build system

Andarist:
@Pomax it wont fix this, the only difference between `main` and `jsnext:main`/`module` entries is what type of modules they might contain, latter may contain es6 import/export statements (and those will be recognized and transpiled by the modern bundler like webpack2 and rollup)

The rest of their content (other es6 features like `let`, `const`, `class`) wont get automatically transpiled.

As I need to ship es5 code to the browser I've setup my webpack2 with such config, so this lib gets transpiled too:
```js
{test: /\.jsx?/, use: ['babel-loader'], exclude: /node_modules\/(?!react-onclickoutside)/},
```

Pomax:
@Andarist might be worth putting that in the README.md as part of the installation and use section, under a "webpack + babel" subheading.

Dante-101:
@Pomax  and @Andarist, Thanks for the super quick replies! :)

`exclude: /node_modules\/(?!react-onclickoutside)/}` works but changing webpack configuration for one package is not a desirable situation.

It will be great if sometime down the line we can release a version that doesn't need this hack and it works out of the box.

Pomax:
It would be. Ideally, webpack/babel can simply autodetect that a dependency is ES6 and auto-compile it for you.

joeduncan:
@Pomax what's the downside of releasing this package in es5? You can still write your package in es6 and then release both the transpiled and original version without problems. This way users of this package would not have to change their webpack config and it would work out of the box.

thebuilder:
Adding the js:next/module field to package.json seems like a fine solution - But you'd still need to add the es5 compiled .js file. Right now 6.01 only contains `index.js` which all of the entries point to?

[...]

PR #194, opened 2017-05-31
build commonjs and es versions

joeduncan:
As far as i understand the `module` field should point to a version of the package with all es20XX features transpiled but still using modules - so `import` and `export` statements stay in this version. The `main` field should point to fully transpiled version of this package (no es20XX features and no imports/exports).
The `jsnext:main` seems to be [deprecated](https://github.com/rollup/rollup/wiki/jsnext:main
) but its seems reasonable to leave it in for compat reasons.

https://github.com/rollup/rollup/wiki/pkg.module

The module version is transpiled to `/es`, the commonjs to `/lib`.
The folder naming convention is from [redux](https://github.com/reactjs/redux) since it seems to be understandable

fixes https://github.com/Pomax/react-onclickoutside/issues/185


genepaul:
@joeduncan, I just pulled down your fork and tried to `npm run build` and it's complaining about `BABEL_ENV=es`. Did you intend to set the environment variable? `cross-env` is one way to do that. Or am I misunderstanding how this should be run?

joeduncan:
Argh yes I forgot to use cross-env thanks! Are you on windows?

On 1 Jun 2017 5:39 p.m., "Gene Arch" <notifications@github.com> wrote:

@joeduncan <https://github.com/joeduncan>, I just pulled down your fork and
tried to npm run build and it's complaining about BABEL_ENV=es. Did you
intend to set the environment variable? cross-env is one way to do that. Or
am I misunderstanding how this should be run?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/Pomax/react-onclickoutside/pull/194#issuecomment-305532863>,
or mute the thread

genepaul:
Yes I am. It occurs to me that it really doesn't matter if other people can run the builds on other platforms, so that might not be necessary for this. It only matters if the owner can generate the build.

joeduncan:
I will still add cross-env though - building should ideally work out of the
box for everybody :)

joeduncan:
@genepaul cross-env added - please give it a try and tell me if it works :)

genepaul:
Works great @joeduncan! Thanks!

.babelrc
```diff
@@ -0,0 +1,18 @@
+{
```
Pomax: can we put this in `package.json`? babel will pick up on it there.

package.json
```diff
@@ -2,11 +2,13 @@
   "name": "react-onclickoutside",
   "version": "6.0.0",
   "description": "An onClickOutside wrapper for React components",
-  "main": "./index.js",
-  "module": "./index.js",
-  "jsnext:main": "./index.js",
+  "main": "lib/index.js",
```
Pomax: I'd like to keep the es6 version just "index.js". It might get put in a different dir for ES5 purposes but I don't want to put the authoritative file itself in some dir.

package.json
```diff
@@ -2,11 +2,13 @@
...
+  "module": "es/index.js",
+  "jsnext:main": "es/index.js",
   "files": [
-    "index.js"
+    "es",
+    "lib",
+    "dist"
```
Pomax: what is `dist` for?

.babelrc
genepaul: This actually causes problems up the build chain if you're using an older version of babel (5). I realize babel 5 is deprecated, but there are those still using it (like my organization). When I tried to babelify this module, it complained about the babel config in the package.json. If you put the config in `.babelrc` then don't include that in the npm package (via `files`), it plays nicely with consumers.

joeduncan: @Pomax should i still move it to `package.json`?

Pomax: I don't particularly like supporting old versions of anything, but in this case it's not a big enough difference to matter. Let's keep it as its own file for now. When babel 7 is out, that'll be the time to go "yeah no seriously stop using babel 5".

joeduncan:
@Pomax is there anything more I can do here? From my side this is ready to ship :shipit:

Pomax:
I'm running it through a test pass and if it checks out I'll land it with a new minor version

Pomax:
v6.1.0 published

joeduncan:
@Pomax thanks!

Pomax:
and you!

Andarist:
@Pomax 
Not being rude or anything - just curious. What made u to change or strong stand about shipping es6? As u remember we've argued a little bit about this during my original PR with es6 class refactor and u were quite inconvincible 😃

Pomax:
@Andarist perfectly reasonable question: it's mostly about demand. I don't want to do this, I hate the very notion, but when you suggested it, it was just you, whereas now it's quite a number of people and there is a clear sign that the users want it, which overrules my hatred for still having ES5 code in a repo.

Issue #197, opened 2017-06-08
add a ToC to the readme.


Pomax:
It would certainly make it easier to find "what's in the REAMDE at all"

Andarist:
@Pomax this can be closed as its done and released :)

Pomax:
haha, good point.

PR #200, opened 2017-06-19
Add TOC

GeKorm:
Addresses #197

Added this manually but there are plenty of automated solutions, they come as packages, editor plugins, or online converters.


Pomax:
I don't think there's a strong enough case here to justify tooling when the readme itself is a manual affair - thanks for the help, in it goes!

PR #201, opened 2017-06-29
Exporting IGNORE_CLASS_NAME constant, so people dont have to hardcode…

### LINKED ISSUE #198: Export the default 'ignore' classname as a constant
According to the docs,
> the default string used is `ignore-react-onclickoutside`

It would be nice to be able to import this class name as a constant in my code instead of relying on hardcoded strings.

What do you think?

Andarist:
… it in their projects (fixes #198)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,6 +8,9 @@ import generateOutsideCheck from './generateOutsideCheck';
 const registeredComponents = [];
 const handlers = [];
 
+const touchEvents = ['touchstart', 'touchmove'];
+export const IGNORE_CLASS_NAME = 'ignore-react-onclickoutside';
+
 /**
  * This function generates the HOC function that you'll use
  * in order to impart onOutsideClick listening to an
@@ -22,7 +25,7 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
     static defaultProps = {
       eventTypes: ['mousedown', 'touchstart'],
       excludeScrollbar: (config && config.excludeScrollbar) || false,
-      outsideClickIgnoreClass: 'ignore-react-onclickoutside',
+      outsideClickIgnoreClass: IGNORE_CLASS_NAME,
       preventDefault: false,
       stopPropagation: false,
     };
@@ -129,10 +132,12 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
         if (!events.forEach) {
           events = [events];
         }
+
         events.forEach(eventName => {
-          const handlerOptions = !this.props.preventDefault && ['touchstart', 'touchmove'].indexOf(eventName) !== -1
-            ? { passive: true }
-            : null;
+          let handlerOptions = null;
+          if (!this.props.preventDefault && touchEvents.indexOf(eventName) !== -1) {
+            handlerOptions = { passive: true };
+          }
           document.addEventListener(eventName, fn, handlerOptions);
         });
       }


src/index.js
```diff
@@ -130,9 +132,10 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
           events = [events];
         }
         events.forEach(eventName => {
-          const handlerOptions = !this.props.preventDefault && ['touchstart', 'touchmove'].indexOf(eventName) !== -1
-            ? { passive: true }
-            : null;
+          const handlerOptions =
```
Andarist: thats on `prettier` ;)

Pomax: there is nothing pretty about this tertiary. Instead of making it "an object or null", let's do this instead:
```
const handlerOptions = {
  passive: false
};

const touchEvents = [
  'touchstart',
  'touchmove'
];

if (!this.props.preventDefault && touchEvents .indexOf(eventName) > -1) {
  handlerOptions.passive = true;
}
```

Andarist: this ternary is used here as a workaround for phantomjs quirk https://github.com/ariya/phantomjs/issues/14982 (from what I remember)

Andarist:
Ammended with `touchEvents` array extracted

src/index.js
Pomax: can you also update the `handlerOptions` though? it's much cleaner as code that always sets up an object.

Andarist: as mentioned - it breaks phantomjs tests:
- should not call handleClickOutside if this.props.disableOnClickOutside() is called, until this.props.enableOnClickOutside() is called.
- should attach and deattach event listener on updates
- should handle disabling and enabling onClickOutside listener when having no DOM node

As u can see all of them involves deattaching the DOM listener. To make it work we would have to store the reference to the object used while using `addEventListener` to pass it to `removeEventListener` later - solely to support phantomjs' quirk as its working just fine in real browsers.

Pomax: You're going to have to explain this a bit more then, because those three things will be true irrespective of whether you use a ternary here to turn `handlerOptions` into an "object or a null", or whether we use proper code to make sure it's always an object, just with `passive` set appropriately.

In both cases, that event listening options object will need to be available (either because it got cached or because we regenerate it) when `addEventListener` is invoked. Changing the assignment for `handlerOptions` to be an object with `{ passive: false }` rather than a `null` by default should make zero difference in execution with respect to addEventListener/removeEventListener functionality

Andarist: Yes, all you have said is true - if we follow the standard and all. However I suspect that PhantomJS (as its implementing the standard) introduced a bug and it stores all arguments references used in the `addEventListener` call to compare them with equality check later when executing `removeEventListener` call.

I've filed an issue on the PhantomJS for this, but no answer - https://github.com/ariya/phantomjs/issues/14982

Pomax: Then the solution, honestly, is "let's stop using Phantom because it's not compatible with the modern web". I know firefox recently added a headless feature, perhaps we can use that.

Andarist: Chrome did release headless mode too recently. Anyway - switching away from phantom is beyond the scope of this PR.

Pomax: Given the fact that this is a convenience PR, I'm okay with blocking this on updating first. We can see if switching to Slimer as a first step lets us test with the desired code.

Pomax:
currently blocked on https://github.com/Pomax/react-onclickoutside/issues/208

Andarist:
@Pomax I think this shouldnt be blocked on #208 . The thing you want to solve with #208 is totally not part of this PR and is already on master and npm. This PR is a simple change and focuses only on exporting `IGNORE_CLASS_NAME`.

Pomax:
@Andarist as a non-blocking improvement I don't want to land this code with that tertiary still in there. At the very least let's rewrite this to a default and override-when-needed:
```
var handlerOptions = null;

if (!this.props.preventDefault && touchEvents.indexOf(eventName) !== -1) {
  handlerOptions = { passive: true };
}
```

Andarist:
@Pomax ammended with the proposed change

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
+npm-debug.log
```
Pomax: I always leave this out - if your commit contains an npm-debug.log, you need to fix that before pushing your commit, not commit despite there being a npm failure somewhere. Seeing that file is a very strong signal that something is wrong and you should not be pushing yet.

src/index.js
```diff
@@ -129,10 +131,14 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
         if (!events.forEach) {
           events = [events];
         }
+
+        const touchEvents = ['touchstart', 'touchmove'];
```
Pomax: as a constant, declaring it in this scope means it'll get redeclared despite never changing. Let's move this up to where `IGNORE_CLASS_NAME` gets declared (but without exporting it)

Andarist:
@Pomax ammended with changes

Pomax:
yeah that works for me. Merging.

Pomax:
I had to back this out - by adding more than just the default export, the library requirement changes from "exporting a function" to "exporting an object with a property `.default` that is the HOC". This is a breaking change, and cannot go into v6.*

PR #202, opened 2017-06-29
Deps cleanup

Andarist:
Last thing to do is to establish what should be done with rollup deps. They were left out from the es6 refactor, but afterwards the build chain has changed and it doesnt use it anymore - just pure babel.

We have 2 options:
- remove them
- leverage rollup to provide a "flat bundle" - `lib` and `es` directories could contain just a single file in them respectively, advantage is that any internal is hidden and its only possible to import whats public + better babel helpers hoisting (no gain now, as `generateOutsideCheck.js` doesnt contain any - but probably an auto gain for the future)


Pomax:
Nice. Can you file an issue that we can tie this PR to, and where we can discuss the needs/merits/etc?

package.json
```diff
@@ -29,15 +29,15 @@
...
   },
   "scripts": {
-    "build": "npm run build:es && npm run build:commonjs",
+    "build": "run-p build:*",
     "build:es": "cross-env BABEL_ENV=es babel src -d es",
     "build:commonjs": "cross-env BABEL_ENV=commonjs babel src -d lib",
     "lint": "eslint src/*.js ./test",
     "test": "run-s test:**",
-    "test:basic": "run-s lint build",
+    "test:basic": "run-s lint build:commonjs",
```
Pomax: in the interest of proper coverage, we should also add a `test:es` here that tests the pure "modern ES" version.

package.json
```diff
@@ -29,15 +29,15 @@
...
     "test:karma": "karma start test/karma.conf.js --single-run",
     "test:nodom": "mocha test/no-dom-test.js",
-    "precommit": "npm test && lint-staged",
+    "precommit": "lint-staged && npm test",
```
Pomax: This seems the wrong order: `npm test` runs `test:**`, which runs `test:basic` which in turn runs `run-s lint build:commonjs` so running `lint-staged` first means that it's first linting, then building.

package.json
Andarist: yeah, probably we should, cannot spare time now to configure it though :/ also it only would test other module format, no other code changes, `es` build is all es5 except of `import`/`export`syntax

Pomax: It would, however, test something we can't otherwise test: the conversion to old ES. While it is reasonable to assume babel has its own test suite, I've seen enough odd conversion quirks in the past to justify testing both just in case the original passes and the other fails.

I don't think it'll need any special affordances though: phantom should support modern JS... but as long as we file an issue for it we can deal with that on its own instead of in this PR

package.json
Andarist: hm, u are quite right, its a duplication (as it already was :D)

now its `lint-staged -> lint (all) -> build:commonjs -> run tests`
before it was - `lint (all) -> build:commonjs -> run tests -> lint-staged`

I've changed the order as I wanted to format the changed files first and THEN run tests as reformatting code might lead (in extreme situation, this should never happen!) to some unexpected changes. So it seemed reasonable to run tests after they get reformatted.

Personally i would remove linting from `test` task, those are imho 2 different things anyway. Linting errors still will be catched before commiting thanks to `lint-staged`

Pomax: "test" is your general runner for everything from linting to unit testing to coverage, etc, so removing linting from `test` is not really a good idea. However, whatever changes `npm test` introduces due to the build step updates files that are probably already in the added changeset, so putting lint-staged last checks against the updated files, which is better than checking first and then rebuilding.

package.json
Pomax: I've filed #206 for the additional npm script.

package.json
Andarist: ok, ive reversed the change

Pomax:
in it goes

Pomax:
I had to back this out - by adding more than just the default export, the library requirement changes from "exporting a function" to "exporting an object with a property `.default` that is the HOC". This is a breaking change, and cannot go into v6.*

PR #205, opened 2017-06-29
Refactored internal quite a lot - simplified the code, stripped down …

### LINKED ISSUE #159: onClickOutside doesn't respond to Composed component's updated props

Andarist:
…attachment/deattachment of the handler logic (in favor of simple in handler checks). It now doesnt keep stale handler (with encapsulated variables) attached to the document (fixes #159)

Still needs to write a test for the fixed issue. Before that though please look at the PR. There are quite a lot of changes (mostly deleted/moved code), so this needs a proper review.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,12 +1,9 @@
 import { createElement, Component } from 'react';
 import { findDOMNode } from 'react-dom';
-import generateOutsideCheck from './generateOutsideCheck';
+import * as DOMHelpers from './dom-helpers';
+import uid from './uid';
 
-/**
- * A higher-order-component for handling onClickOutside for React components.
- */
-const registeredComponents = [];
-const handlers = [];
+const handlersMap = {};
 
 const touchEvents = ['touchstart', 'touchmove'];
 export const IGNORE_CLASS_NAME = 'ignore-react-onclickoutside';
@@ -43,14 +40,35 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
       return ref.getInstance ? ref.getInstance() : ref;
     }
 
-    // this is given meaning in componentDidMount/componentDidUpdate
-    __outsideClickHandler = null;
+    __outsideClickHandler = event => {
+      if (typeof this.__clickOutsideHandlerProp === 'function') {
+        this.__clickOutsideHandlerProp(event);
+        return;
+      }
+
+      const instance = this.getInstance();
+
+      if (typeof instance.props.handleClickOutside === 'function') {
+        instance.props.handleClickOutside(event);
+        return;
+      }
+
+      if (typeof instance.handleClickOutside === 'function') {
+        instance.handleClickOutside(event);
+        return;
+      }
+
+      throw new Error(
+        'WrappedComponent lacks a handleClickOutside(event) function for processing outside click events.',
+      );
+    };
 
     /**
      * Add click listeners to the current document,
      * linked to this component's state.
      */
     componentDidMount() {
+      this._uid = uid();
       // If we are in an environment without a DOM such
       // as shallow rendering or snapshots then we exit
       // early to prevent any unhandled errors being thrown.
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -40,8 +38,31 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
       return ref.getInstance ? ref.getInstance() : ref;
     }
 
-    // this is given meaning in componentDidMount/componentDidUpdate
-    __outsideClickHandler = null;
+    // this is given meaning in enableOnClickOutside
+    __internalOutsideClickHandler = null;
```
Andarist: wanted to split those 2 - `__internalOutsideClickHandler` and `__outsideClickHandler` but had not better names in my head 😄

Andarist:
@Pomax thats something I would like to push forward, ur code review would be highly appreciated - I'll stall this for a little bit longer, before I decide to give it a go and merge in.

src/index.js
```diff
@@ -43,14 +40,35 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
       return ref.getInstance ? ref.getInstance() : ref;
     }
 
-    // this is given meaning in componentDidMount/componentDidUpdate
-    __outsideClickHandler = null;
+    __outsideClickHandler = event => {
+      if (typeof this.__clickOutsideHandlerProp === 'function') {
```
Pomax: does this need a typeof check? It feels like if this property exist, we can reasonably assume it's going to be a function. Given that this will now be checked continuously, making this as fast as it reasonably can be feels worth it:
```
__outsideClickHandler = event => {
  if (this.__clickOutsideHandlerProp) {
    return this.__clickOutsideHandlerProp(event);
  }

  ...
}
```

(with the same updates for the subsequent checks)

src/uid.js
```diff
@@ -0,0 +1,5 @@
+function autoInc(seed = 0) {
```
Pomax: does babel understand generators? If so we could rewrite this as:
```
function* autoInc() {
  var uid = 0;
  while (true)
    yield uid++;
}
```
rather than exporting a function that returns a function.

Pomax:
as long as we're still passing tests and the API didn't change, I'm fine with this (barring the few PR comments). If this changes the API, we'll probably want to tag this (once merged) as the next major version

src/index.js
Andarist: Yeah, we need to only check for existance. I just have picked check against function as its explicit and probably not more costful than any other type of check (probably even better than smth like if (prop) as each coerced possibilities need to be checked)

Do u feel this needs to be changed? I assume this is a fast path as like ive mentioned it performs only a single check - instead guarding against nulls, undefineds, 0s, empty strings and whatnot

src/uid.js
Andarist: Babel understands generator and can transpile them just fine but transpiled version is not nice - its a machine state which has switch statement in its body for each state which can be understood by regenerator-runtime and its way too big for this simple case unfortunately.

U can check transpiled generators in babels online repl and check out regenerator-runtime's code in their repo

Andarist:
This is not a breaking change. At least it shouldnt be ;) tests are passing and i dont expect it to break anyone - but well, we never know that for sure :P

src/index.js
Andarist: What ive written still stands, although i now see that u have proposed a simpler (for human) check which needs to apply all coercion rules. Havent noticed it before as im answering from my phone, so my comment might have sounded a little bit disconnected from ur comment. Anyway - like ive said, i believe this is faster (function check) in modern engines, although ofc this is micro optimization anyway

Andarist: https://docs.google.com/presentation/d/1_eLlVzcj94_G4r9j9d_Lj5HRKFnq6jgpuPJtnmIBs88/edit#slide=id.g2134da681e_0_596

I wouldnt expect `typeof fn === 'function'` being much different from `obj !== undefined`, but can change to the latter

src/uid.js
Andarist: https://babeljs.io/repl/#?babili=false&browsers=ie%2011&build=&builtIns=false&circleciRepo=&code_lz=GYVwdgxgLglg9mAVAAgIYinAkpAFASmQG8AoZZAN1QCdkQYATZAXmQAYBuM5AdwAsYAGwCmyXFGohh-buQCeMYYKb0GAajVcAvkA&debug=false&evaluate=false&lineWrap=false&presets=es2015&prettier=false&showSidebar=true&targets=&version=6.26.0

+ this needs to be included for it to work - https://github.com/facebook/regenerator/blob/30d34536b9e3f7a2873b04a16ec66fec9c8246f6/packages/regenerator-runtime/runtime.js

PR #213, opened 2017-07-27
Fixed examples of requiring the library with (fixes #211)

### LINKED ISSUE #211: Update documentation
The example should be updated to:
`const OnClickOutside = require('react-onclickoutside').default`

Andarist:


Andarist:
@Pomax please check out if now its better, ive refactored all examples to es6 classes and added a note about `.default` in 1 place where usage with `require` is shown now

Issue #210, opened 2017-07-10
Error with eslint no-unresolved since 6.2.0


tofumatt:
See the error here: https://travis-ci.org/mozilla/addons-frontend/jobs/252183587#L1977

Since `6.2.0` I get the following error when running eslint against my repo (addons-frontend): `Unable to resolve path to module 'react-onclickoutside'  import/no-unresolved`

This happens for `6.3.x` as well. Not sure what caused it but it isn't happening elsewhere, wondering if there was a recent refactor/change of file structure that would've caused it.

Pomax:
that's very odd, as 6.3 should have been a rollback to 6.1... let me try to re-rollback O_o

Pomax:
@Andarist any idea?

Pomax:
I've published 6.4, which should be a reversion to 6.1.1 - let me know if that's fixed things @tofumatt

tofumatt:
[Passed all my tests](https://github.com/mozilla/addons-frontend/pull/2741). Thanks!

Andarist:
@Pomax @tofumatt 
That is turbo strange o.O @tofumatt would u be able to provide a repository with reproduced issue? Would be helpful for debugging purposes.

tofumatt:
Indeed, the repo is [addons-frontend](https://github.com/mozilla/addons-frontend/). Here are the relevant failing PRs:

* https://github.com/mozilla/addons-frontend/pull/2731 (`6.2.0`, possibly unrelated)
* https://github.com/mozilla/addons-frontend/pull/2732 (`6.3.0`)
* https://github.com/mozilla/addons-frontend/pull/2733 (`6.3.1`)

Andarist:
Thanks, gonna set it up in a minute !

Andarist:
Ok, I know what happened. `es` directory was not published on those versions, and the resolver (rightfully) tried to access it (as specified in `package.json`'s `module` entry).

@Pomax prepublish script should be added to ensure valid publications

Pomax:
@Andarist whatever was the latest node 7.x with npm 3.x - I updated it since due to the Nodejs security advisory since, so maybe npm3 was too low and it needed 4 or 5?

Andarist:
Hm, aint sure how to properly maintain `scripts` for various version of npm. there might not be any, but as only u have publishing rights and u have updated ur node and npm (from what I understand) - this shouldnt happen anymore.

Anyway - republishing what got reverted should be made I think.

[...]

PR #214, opened 2017-07-27
add minimal node and npm requirements

Pomax:


Andarist:
Please see my comment [here](https://github.com/Pomax/react-onclickoutside/issues/210#issuecomment-318601717)

Andarist:
As per https://docs.npmjs.com/files/package.json#engines
> You can also use the "engines" field to specify which versions of npm are capable of properly installing your program.

It wouldnt protect us much and only warns users of the package needlessly. If we want to protect ourselves against using older node/npm during publication we need to incorporate other check. I've prepared it and created a new PR - https://github.com/Pomax/react-onclickoutside/pull/224

PR #215, opened 2017-07-28
Allow React 16 as peer dependency

Andarist:


Pomax:
neat, I did not know that was an option.

PR #226, opened 2017-08-28
allow multiple ignore classes

hzuo:
Only allowing a single ignore class has two issues:
1. Forces you to pollute your entire app with that ignore class. You should be able to declare at the site of the target component all the classes you want to blacklist.
2. In many cases there are plugin architectures where there are elements whose classes you can't control (e.g. popovers) that you still want ignored.

### DIFF
--- a/src/generateOutsideCheck.js
+++ b/src/generateOutsideCheck.js
@@ -1,4 +1,21 @@
 /**
+ * Check if classList contains any of the given classes,
+ * transparently handling singular and array cases.
+ */
+function containsAny(classList, classes) {
+  if (Array.isArray(classes)) {
+    for (const className of classes) {
+      if (classList.contains(className)) {
+        return true;
+      }
+    }
+    return false;
+  } else {
+    return classList.contains(classes);
+  }
+}
+
+/**
  * Check whether some DOM node is our Component's node.
  */
 function isNodeFound(current, componentNode, ignoreClass) {
@@ -12,9 +29,9 @@ function isNodeFound(current, componentNode, ignoreClass) {
   // See: http://www.w3.org/TR/SVG11/struct.html#InterfaceSVGUseElement
   // Discussion: https://github.com/Pomax/react-onclickoutside/pull/17
   if (current.correspondingElement) {
-    return current.correspondingElement.classList.contains(ignoreClass);
+    return containsAny(current.correspondingElement.classList, ignoreClass);
   }
-  return current.classList.contains(ignoreClass);
+  return containsAny(current.classList, ignoreClass);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/generateOutsideCheck.js
```diff
@@ -1,4 +1,21 @@
 /**
+ * Check if classList contains any of the given classes,
+ * transparently handling singular and array cases.
+ */
+function containsAny(classList, classes) {
+  if (Array.isArray(classes)) {
```
Andarist: would be better to just pack in an array if its not one yet and handle only one case below that

src/generateOutsideCheck.js
```diff
@@ -1,4 +1,21 @@
...
+    for (const className of classes) {
```
Andarist: `for-of` loop could be avoided here (its creating quite [weirdo code after transpilation](https://babeljs.io/repl/#?babili=false&browsers=ie%2011&build=&builtIns=false&circleciRepo=&code_lz=GYVwdgxgLglg9mABBBUCGMwGcCCYCeAFBADZpZYAyMWUANMmRQKZYCUiA3gFCKIzBEhHACcRafADoao8UVLksrNhx58-wOCKEpsURooByaALbNEcQQpbsuvdXwE6mVGlEm70mLMRfGzKnYODiLMUCAiSFAiIMwA3PYOAL6JiCkhYRFIwGgkSgl8SYjMeeZq6qHhkQYU1LQeqBjYvorKBWncSUA&debug=false&evaluate=false&lineWrap=false&presets=es2015&prettier=false&showSidebar=true&targets=&version=6.26.0), although we use `loose` mode and its slightly better than what I have presented) and it could be handled by easy `Array.prototype.some` call

hzuo:
Cool - generalized containment function definitely makes sense. Should I try to make the above changes to get this merged or should we just wait on your refactor? The current PR is probably good for people who want the new functionality without sacrificing back-compat

hzuo:
@Andarist

Andarist:
@hzuo Im still not sure if I want to merge this, requested changes after code review are really rather cosmetic, so you might try to adjust the code accordingly. No promises about merging it though, but I won't say definite "no" about this at the moment.

Andarist:
@hzuo sorry for a delay - I've created a PR i was planning to do for quite some time. I'd love your input there - https://github.com/Pomax/react-onclickoutside/pull/246

augbog:
Hey just to clarify is this PR blocked because of #246? I'd be interested in this feature :)

Andarist:
@augbog yeah, i need to finish the #246 (wanna help with that? 😉 ), but got occupied with tons of other OSS work and somehow never got back to it :s

driplo:
Hello, any news on this PR ?

Pomax:
I would prefer not to land this PR in favour of a containment function, but I would also advocate adding that containment function independent of any overhaul. As long as the API stays the same across both changes, no one will notice the rug got replaced under their feet.

PR #227, opened 2017-09-13
Add config options for all default props

penx:
Allow excludeScrollbar, preventDefault, outsideClickIgnoreClass and stopPropagation props to be specified in config

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -23,11 +23,11 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
     static displayName = `OnClickOutside(${WrappedComponent.displayName || WrappedComponent.name || 'Component'})`;
 
     static defaultProps = {
-      eventTypes: ['mousedown', 'touchstart'],
+      eventTypes: (config && config.eventTypes) || ['mousedown', 'touchstart'],
       excludeScrollbar: (config && config.excludeScrollbar) || false,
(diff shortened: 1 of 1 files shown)


Andarist:
Makes sense :) Thanks for the contribution. Could add some simple tests for this PR?

bpartridge:
As a workaround while this is being tested, it's relatively simple to modify the class returned by onClickOutside(...) to change its default props; as with something set in config, they'd still be overridden by anything passed in by the parent.

```js
const ClickOutsideWrapped = onClickOutside(MyComponent);
ClickOutsideWrapped.defaultProps.eventTypes = ['touchstart']; // default to only on mobile
export default ClickOutsideWrapped;
```

Either way, if something high-level i.e. something that would be loaded into a section in react-router is wrapped with this, this workflow allows us to specify eventTypes, etc. behavior in the component's own file, rather than splitting that logic between the owner and the component.

lcsrinaldi:
What's the progress with this PR? I've opened an issue that would be solved by it. https://github.com/Pomax/react-onclickoutside/issues/265

penx:
I may get around to writing some tests at some point, but as I'm not currently using react-onclickoutside this is unlikely to be any time soon. Anyone's welcome to fork my changes and create a new PR with tests.

PR #230, opened 2017-09-27
build an UMD module

### LINKED ISSUE #207: Rollup dependency cleanup
Its a followup for #203

>There are still rollup dependencies leftovers in the `package.json`. It should be decided if they should be removed or kept and being used for something.
> It can still be valuable to leave it to:
> - provide flat bundle (single file bundle - keeping internals hidden)
> - build UMD version of the lib, so its easy to include on the webpage and experiment

jochenberger:
fixes #207


.babelrc
```diff
@@ -1,5 +1,17 @@
 {
     "env": {
+        "umd": {
```
Andarist: new env is not rly needed, just re-use `es`

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 lib
 es
+/dist/
```
Andarist: I prefer not having this in `.gitignore` - its often nice to look on the visual diff in the umd bundles, especially when updating deps etc

rollup.config.js
```diff
@@ -0,0 +1,25 @@
+import nodeResolve from 'rollup-plugin-node-resolve';
+import babel from 'rollup-plugin-babel';
+
+var config = {
```
Andarist: u can use `const` here, this file gets transformed with babel when loaded (rollup does it internally)

rollup.config.js
```diff
@@ -0,0 +1,25 @@
...
+    nodeResolve({
+      jsnext: true
+    }),
+    babel({
+      exclude: 'node_modules/**',
+    })
+  ]
+};
+
+export default config;
```
Andarist: Please update rollup (and other rollup-related dependencies) to its latest version. Rollup's tree-shaking algorithm has changed a little bit and for now I advise to filter out `asyncGenerator` helper like [here](https://github.com/redux-saga/redux-saga/blob/9382f74c3e490fb952ae6f8fe71e52fbc5609f97/rollup.config.js#L21)

The linked config was recently upgraded, you can find the commit and see what's needed by the new rollup API

rollup.config.js
```diff
@@ -0,0 +1,25 @@
...
+  globals: {
+    react: 'React',
+    'react-dom': 'ReactDOM'
+  },
+  plugins: [
+    nodeResolve({
+      jsnext: true
+    }),
+    babel({
+      exclude: 'node_modules/**',
```
Andarist: please use `external-helpers` plugin

package.json
```diff
@@ -32,6 +33,7 @@
     "build": "run-p build:*",
     "build:es": "cross-env BABEL_ENV=es babel src -d es",
     "build:commonjs": "cross-env BABEL_ENV=commonjs babel src -d lib",
+    "build:umd": "cross-env BABEL_ENV=umd rollup -c",
```
Andarist: it would be great to provide 2 umd builds, second one would be minified (based on `NODE_ENV=production/development` - this will require adding uglifyjs plugin and using it conditionally in the `rollup.config.js`)

.gitignore
Andarist: it would also be helpful now to see what bundle got produced by this PR

rollup.config.js
```diff
@@ -0,0 +1,25 @@
+import nodeResolve from 'rollup-plugin-node-resolve';
```
Andarist: aint sure, but this plugin might not be needed here, we do not import any dependencies from the `node_modules` to be included in the bundle (`react` and `react-dom` are treated as external)

so would be nice to have it checked and possibly removed

rollup.config.js
```diff
@@ -0,0 +1,25 @@
...
+import babel from 'rollup-plugin-babel';
+
+var config = {
+  entry: 'src/index.js',
+  dest: 'dist/ReactOnClickOutside.js',
```
Andarist: would prefer `dist/react-onclickoutside.js` so it reflects npm package name

package.json
```diff
@@ -7,7 +7,8 @@
   "jsnext:main": "es/index.js",
   "files": [
     "es",
-    "lib"
+    "lib",
+    "dist"
   ],
```
Andarist: we can add `unpkg: "dist/react-onclickoutside.min.js"`, so it would produce nicer links for the unpkg service

Andarist:
**Note**: This PR is focused on getting UMD bundle, but rollup is also great for 'flat bundles'. This could be done too, but probably as a followup PR once this one gets merged in. I'll probably take care of it later.

.babelrc
jochenberger: Alright, I just thought it's cleaner this way.

rollup.config.js
jochenberger: Will do. I just restored the old config from the git history.

rollup.config.js
jochenberger: Shouldn't we do that via a separate babel env then?

.babelrc
Andarist: i see `env` here as specifying needed babel transforms, not the target (output type) per se

rollup.config.js
Andarist: that is certainly approach that could be done, but `external-helpers` in rollup's context are somewhat magical and that magic is tightly coupled with rollup, so I think its a good place to have it the rollup's config rather than `.babelrc`

doesnt matter so much anyway, just got used to doing it like this, its also what most libs ive seen do

dist/react-onclickoutside.js
```diff
@@ -0,0 +1,320 @@
...
+    outsideClickIgnoreClass: IGNORE_CLASS_NAME,
+    preventDefault: false,
+    stopPropagation: false
+  }, _class.getClass = function () {
+    return WrappedComponent.getClass ? WrappedComponent.getClass() : WrappedComponent;
+  }, _temp2;
+}
+
+exports.IGNORE_CLASS_NAME = IGNORE_CLASS_NAME;
+exports['default'] = onClickOutsideHOC;
```
Andarist: when you load it, remember to use `window. onClickOutside.default`

rollup.config.js
jochenberger: The plugin increased the bundle size from 11945 to 12048.

package.json
jochenberger: Didn't know that one. 👍

rollup.config.js
```diff
@@ -0,0 +1,51 @@
+import babel from 'rollup-plugin-babel';
+import { list as babelHelpersList } from 'babel-helpers';
+import uglify from 'rollup-plugin-uglify';
+import { minify } from 'uglify-es';
```
Andarist: not so sure that we should pass custom uglify to the plugin, `uglify-es` is not yet considered stable enough and `rollup-plugin-uglify` has its dependency on `uglify-js@3`

jochenberger:
Should be fine now. Do you want be to squash the commits into one?

Andarist:
I would appreciate squashing, although nowadays I can also do it myself easily choosing Squash and merge option in GitHub's UI.

Andarist:
Sorry, didnt yet have time to review and merge, gonna do it asap when I have time.

PR #245, opened 2017-11-26
Niv/expand exclude scrollbar functionality

nivos-saar:
- When the user chooses to exclude the scrollbar clicks, it affects only on the browser scrollbar.
- There are many cases that a web application doesn't use the browser scrollbar, then the scrollbars you'll see in the app are overflows of different containers.
- I expanded the clickedScrollbar functionality to check if the target of the user's click is in a container with scrollbar, if so it's check his X & Y to see if it's on the scrollbar or somewhere else.


Andarist:
I've created a PR i was planning to do for quite some time. I'd love your input there - https://github.com/Pomax/react-onclickoutside/pull/246

nivos-saar:
@Andarist Hey, it's been more than a month. Can you please Approve, I prefer not to fork it.

OmriAharon:
Hey @Andarist!

This PR has been open for 4 months, can you please share your intentions?
If there's no activity here we'll just fork to our own.. But otherwise this is not a big change so would be great to see it get in.

PR #247, opened 2017-12-06
Handle unsupported accessor exception

strayiker:
Not all browsers support accessors (get/set), so it will be good to handle this and keep code is working.

### DIFF
--- a/src/detect-passive-events.js
+++ b/src/detect-passive-events.js
@@ -6,17 +6,20 @@ export const testPassiveEventSupport = () => {
   }
 
   let passive = false;
-
-  const options = Object.defineProperty({}, 'passive', {
-    get() {
-      passive = true;
-    },
-  });
-
   const noop = () => {};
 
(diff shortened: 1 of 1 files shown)


Andarist:
Could u specify what browser does not support it?

strayiker:
IE8 as example )

Andarist:
react@15+ does not have official support for ie8. If your react app works on ie8 by some chance, I'd suggest downgrading `react-onclickoutside` to some previous version (even some of the `6.x` line)

strayiker:
Yes, i know. But is still work on ie8 with some polyfills. And i will be happy to be able to use the latest version of this awesome library for new browsers but with support for older.

Pomax:
IE8 at this point in time is an officially unsupported browser, for an officially  unsupported operating system. We do not need to make provisions for that in the same way there's no need to support IE4 or Android 1.0 Webviews.

PR #252, opened 2018-01-04
add passive event listener options to removeEventListener

### LINKED ISSUE #251: removeEventListener not being called in certain browsers
The call to removeEventListener does not contain any options, and this means it will fail in certain browsers (see the Matching event listeners section of https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener). 

> It's worth noting that some browser releases have been inconsistent on this, and unless you have specific reasons otherwise, it's probably wise to use the same values used for the call to addEventListener() when calling removeEventListener().

This causes many errors once a component is unmounted (Cannot read property 'getInstance' of null).

I'll make a PR to add it.

The browser I have that runs into this issue is a Chrome 30 webview on a samsung tab 4 (i was surprised to see it passed the feature detection, but it does). But it's not all versions of chrome 30, (couldn't reproduce on browserstack). So sorry for being unable to provide a reproduction.

bdwain:
fixes #251


src/index.js
```diff
@@ -110,6 +110,19 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
       this.disableOnClickOutside();
     }
 
+    /**
+     * Options for addEventHandler and removeEventHandler
+     */
+    getEventHandlerOptions(eventName) {
+      let handlerOptions = null;
```
Andarist: Could you just hoist this to the top of the file and accept `instance` as argument? there is no point to keep this on prototype and expose this as "public" API

bdwain: yup! done.

Andarist:
Thanks

bdwain:
cool thanks for merging it. would it be possible to do a release with this change soon?

Andarist:
I've just released new version

bdwain:
thanks!

PR #255, opened 2018-01-11
Ability to define custom document

### LINKED ISSUE #236: Support for iFrame
If the react component is in an iframe, onclickoutside doesnt really work. You have to click on the parent document to trigger `handleClickOutside` 

This module binds all click events to `document`. If you can expose the `document` prop, i can bind events to an iframe document instead. What do you think?

rmdort:
This pull requests fixes #236  .

The issue:
When a react component is rendered in an iframe, you only want to bind `click` events to the iframe document and not the parent page. The plugin didn't allow it to.

Solution:
The user can configure which `document` to bind events to


src/index.js
```diff
@@ -77,6 +77,7 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
      * linked to this component's state.
      */
     componentDidMount() {
+      const document = config && config.getDocument ? config.getDocument(this.getInstance()) : global.document;
```
Pomax: This check should not be necessary every single time, since dom components cannot be moved from one document to another (only within one document), so we can cache this as `this.document = ...` rather than as local `const document` and then use that in the enabling/disabling functions.

rmdort: Noted. Done. 👍

src/index.js
```diff
@@ -93,7 +94,7 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
       // If we are in an environment without a DOM such
       // as shallow rendering or snapshots then we exit
       // early to prevent any unhandled errors being thrown.
-      if (typeof document === 'undefined' || !document.createElement) {
+      if (typeof this.document === 'undefined' || !this.document.createElement) {
```
Pomax: given the assignment in the constructor, is `this.document` being `undefined` possible? It would probably be better to ensure that `this.document` is always a document, to the point where if L52 yields `null` or `undefined`, we assign a duck-typed document-like object that has exactly as much as the HOC needs to at least think it's working, and then warn about the fact that we had to do that with a `console.warn` or `console.error`.

This removes a whole bunch of checks the result of which should already be known at the end of running the constructor (since we can't move components from one document to another, even if we could conceivably serialize/deserialize across documents)

src/index.js
```diff
@@ -49,6 +49,7 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
     constructor(props) {
       super(props);
       this._uid = uid();
+      this.document = config && config.getDocument ? config.getDocument(this.getInstance()) : global.document;
```
Andarist: You could just make `getDocument` having a default value:
```js
const defaultGetDocument = () => typeof document !== "undefined" ? document : null
export default function onClickOutsideHOC(WrappedComponent, {
  getDocument = defaultGetDocument,
  excludeScrollbar = false,
  handleClickOutside,
} = {}) {
```
and remove those ifs (`config &&`) all over the place (well, in like 3 places :P).

It is not a good practice to use `global` in library, there is no such thing and it has to be polyfilled by users to make this work - just try to avoid this if you can.

Does your `getDocument` needs an argument (`this.getInstance()`)? What for?

src/index.js
Andarist: > given the assignment in the constructor, is this.document being undefined possible?

Yeah, it can be undefined (or rather a `null` with my proposed change in the other comment) in case of SSR. We should leave those ifs here.

We could OTOH warn in development mode about returning "broken" document from `getDocument` - we'd have to detect being in a browser though, to skip warning in case of SSR too.

It would look something like this:
```js
const isBrowser = () => typeof document !== "undefined" && document.createElement

// ...

this.document = getDocument(this.getInstance())
if (process.env.NODE_ENV !== 'production' && getDocument !== defaultGetDocument && typeof this.document.addEventListener !== 'function') {
  // or maybe even we should fail fast here with throw new Error?
  console.warn('WRITE A WARNING HERE')
}
```

Andarist: Also this if here:
```js
if (typeof this.document === 'undefined' || !this.document.createElement) {
```
should not be about checking if w have a valid `this.document` but rather about checking if we are in the browser env

src/index.js
rmdort: The reason i have `this.getInstance()` is because only the component knows where its being rendered. Component may have some props, to identify where it should be rendered

Andarist: Seems legit. 👍

rmdort: We are using this for components/widgets that are rendered on iframe using https://github.com/ryanseddon/react-frame-component

Example

````
import listensToClickOutside from '@olasearch/react-onclickoutside'
import Frame from 'react-frame-component'

const Dropdown = listensToClickOutside(() => <div />, {
  getDocument: function (instance) {
    return instance.context.document
  }
})

Dropdown.contextTypes = {
  /* Iframe document */
  document: PropTypes.object
}

const Dropdown = listensToClickOutside(() => <div />, {
  getDocument: function (instance) {
    return instance.context.document
  }
})

ReactDOM.render(
  <Frame>
    <Chat />
  </Frame>
  , document.body
)
````

I want to avoid widgets to bind events to parent document.

Comment (Andarist) [reply src/index.js]: Ok, seems good. Other solution would require using `ref` on `Frame` and retrieving somehow the iframe's document, but it would be too hard to pass that to the HOC's `config`

File: src/index.js (hunk 2 above)
Comment (rmdort) [reply src/index.js]: @Andarist  `isBrowser()` makes sense. So in the constructor, we can just assign document as null if its not a browser env

````
this.document = isBrowser()
        ? config && config.getDocument
          ? config.getDocument(this.getInstance())
          : document
        : null;
````

And all `this.document` checks can be replaced with `isBrowser()` check? Wouldn't that suffice? Test are passing

Comment (Andarist) [reply src/index.js]: yeah, something like that should work, please use default `getDocument` though, will make this clearer (`const getDocument = () => document`)

File: src/index.js [hunk 4]
```diff
@@ -24,6 +24,15 @@ function getEventHandlerOptions(instance, eventName) {
   }
   return handlerOptions;
 }
+/**
+ * Detect if we are in browser env
+ */
+const isBrowser = () => typeof document !== 'undefined' && document.createElement;
+
+/**
+ * Default the default document
```
Andarist: seems like a typo, should just be "Default getDocument" or something like that

src/index.js
```diff
@@ -49,6 +58,9 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
     constructor(props) {
       super(props);
       this._uid = uid();
+      this.document = isBrowser()
+        ? config && config.getDocument ? config.getDocument(this.getInstance()) : getDocument()
```
Andarist: I meant that you should just make a config argument likes this:
```js
(WrappedComponent, {
  getDocument = defaultGetDocument,
  excludeScrollbar = false,
  handleClickOutside,
} = {}) =>
```

and that references to `config` should be just removed from the code.

rmdort: ok. You want me to remove references to config from here as well?

https://github.com/Pomax/react-onclickoutside/blob/master/src/index.js#L102

And change config argument below to default?
````
export default function onClickOutsideHOC(WrappedComponent, config) {
````

rmdort: @Andarist It seems the latest master branch breaks caching `this.document` cos `this.getInstance()`  and ref is only available after componentDidMount. Where can we cache document?

Andarist: You are right, you need to move caching this to the `componentDidMount`

rmdort: Yes, but this test fails

````
it('should register only one click outside listener per instance')

// this.document is not defined
````

When calling `this.props.enableOnClickOutside()` inside componendDidMount

Andarist: Hm, thats a bummer, child's `componentDidMount` gets executed first, so we wont be able to cache it first - problem is that you want to retrieve appropriate document from the instance

I would have to dig deeper into the issue to assess what can be done. Any ideas? Ideally you could pass document in advance (when creating the HOC) instead of reading it from the instance

PR #279, opened 2018-06-25
Fix contains of undefined classList in isNodeFound in IE11

sorenhoyer:
Add a check to make sure classList is not undefined or null, before using contains on it.


Pomax:
I know it's a quick and easy fix, but I still stand behind https://github.com/Pomax/react-onclickoutside#ie-does-not-support-classlist-for-svg-elements

sorenhoyer:
Sorry I should have read the readme. I completely agree with you. Thanks for making me aware of it. :-)

PR #282, opened 2018-07-10
Add support for wrapping arrow function component

cakeseller:
Right now library will produce error if we try to wrap component that was declared with arrow function (since arrow functions doesn't have prototype property).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -55,7 +55,7 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
      * Access the WrappedComponent's instance.
      */
     getInstance() {
-      if (!WrappedComponent.prototype.isReactComponent) {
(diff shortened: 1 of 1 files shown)


Pomax:
Please file an issue that explains this problem first, with some code that illustrates the problem, so that this PR can be tied to an actual problem report as a fixes/closes relation. As is, there isn't enough to warrant review/merging yet.

cakeseller:
Ok, will create an issue with more details as soon as I have free time :)

cakeseller:
I couldn't reproduce issue with minimal required amount of code. Looks like my issue is trickier then I thought and probably it's related to how I define components I my app. Changing component definition from arrow function to class helped me, and I don't have time for further investigation, so I'm closing PR. Sorry for wasting your time.

Pomax:
not wasted: it allowed you to narrow down where the problem was. That's always time well spent!

Issue #281, opened 2018-07-05
Sometimes ref is null or undefined


heygrady:
https://github.com/Pomax/react-onclickoutside/blob/cc1d179bd177615fab0204d0293b771a844068ba/src/index.js#L62

Probably related to https://github.com/Pomax/react-onclickoutside/issues/271

```
TypeError: null is not an object (evaluating 'ref.getInstance')
  File "webpack:///./node_modules/react-onclickoutside/es/index.js", line 136, in getInstance
          return ref.getInstance ? ref.getInstance() : ref;
  File "webpack:///./node_modules/react-onclickoutside/es/index.js", line 43, in __outsideClickHandler
            var instance = _this.getInstance();
  File "webpack:///./node_modules/react-onclickoutside/es/index.js", line 87, in [anonymous]
              _this.__outsideClickHandler(event);
```

We're getting a rollbar issue for this line of code (using version 6.6.2 but the relevant code looks identical to the latest version). We're indirectly including this library because it's a [dependency of react-datetime](https://github.com/YouCanBookMe/react-datetime/blob/master/package.json#L85) (used like [this](https://github.com/YouCanBookMe/react-datetime/blob/master/src/DaysView.js#L6-L9)).

Most of the errors are happening for us in old versions mobile Safari (9).

The precise cause of the issue here is hard to track down. However, it's clear that there are certain circumstances where `this.instanceRef` returns `null` or `undefined`. This is _probably_ due to funky edge cases where the wrapped component is unmounting or renders `null` or is otherwise not an expected value.

```js
getInstance() {
  if (!WrappedComponent.prototype.isReactComponent) {
    return this;
  }
  const ref = this.instanceRef;

  // In some rare circumstances the ref isn't defined or is null. This has been reported
  // to happen in Enzyme and has happened in the wild. It's possible that there are race
  // conditions that occur during mount/unmount that would cause the ref to be undefined.
  // Here we return undefined if we find ourselves in this undefined situation.
  if (!ref) {
    return undefined
  }
  return ref.getInstance ? ref.getInstance() : ref;
}
```

In `__outsideClickHandler`, `componentDidMount` and `componentDidUpdate` we rely on `getInstance` being a real instance. I suggest bailing in those functions if `instance` is not defined.

```js
// ...
const instance = this.getInstance();

// Bail when the instance isn't defined. This would only happen if there is an undefined
// instance ref. That would happen in weird edge cases. Regardless of why, if the ref
[...]

Issue #271, opened 2018-04-08
Enzyme TypeError : Cannot read property 'getInstance' of undefined


Aarbel:
I'm using Enzyme for testing React, which returns this error (i passed from React 15 to 16.3). Do you know a way to avoid it ? 

![image](https://user-images.githubusercontent.com/25119847/38468511-6de17a4c-3b47-11e8-989d-279fae5a9e32.png)

```
TypeError: Cannot read property 'getInstance' of undefined
      at onClickOutside.getInstance (node_modules/react-onclickoutside/dist/react-onclickoutside.cjs.js:272:18)
      at onClickOutside.componentDidMount (node_modules/react-onclickoutside/dist/react-onclickoutside.cjs.js:287:27)
      at node_modules/enzyme/build/ShallowWrapper.js:126:20
      at Object.batchedUpdates (node_modules/enzyme-adapter-react-16/build/ReactSixteenAdapter.js:342:22)
      at new ShallowWrapper (node_modules/enzyme/build/ShallowWrapper.js:125:24)
      at shallow (node_modules/enzyme/build/shallow.js:19:10)
      at Context.<anonymous> (src/components/select/index.test.js:14:9)
```

# To reproduce

**Test file** 
```
import { shallow } from 'enzyme'
import { Select } from '.'

describe('<Select />', () => {

    it('works without options', () => {
        shallow(<Select />)
    })

}
```
**Component**
```
import onClickOutside from 'react-onclickoutside'

class Select extends React.Component {

    render = () => <div></div>
}

const COSelect = onClickOutside(Select)
export { COSelect as Select }
```

Thanks :open_hands:

CoryDuncan:
I'm running into the same `getInstance` error, but not using Enzyme.

Aarbel:
@Pomax ?

CoryDuncan:
I think this might be https://github.com/Pomax/react-onclickoutside/issues/268 because prior to `16.3` it was working as expected.

jigsawye:
@Pomax 
I have same error but not used with enzyme.

In my case, when the component is write with `styled-components`, and call `withOnClickOutside`on the component, the error will throwing.
Because the styled ref is return the `StyledComponent` wrapper, not DOM node.
Now I just wrap a `div` of the component to solved.

Reference: https://www.styled-components.com/docs/advanced#refs
Maybe useful: https://www.styled-components.com/docs/api#isstyledcomponent

Pomax:
[...]

PR #287, opened 2018-08-08
Prevent crash when this.instanceRef is null

teechap:
I can't nail down steps to reproduce, but I'm seeing an issue in a production app where `this.instanceRef` is `null` in `getInstance()`, which causes a crash. I noticed a couple of [open](https://github.com/Pomax/react-onclickoutside/issues/281) [issues](https://github.com/Pomax/react-onclickoutside/issues/271) for this, so I'm sure others are having the same problem. Let's make this more defensive against falsy refs unless there's an obvious breaking change here. :-)


Pomax:
This really should have code that reliably reproduces the error, though. Without that, there is no way to tell whether your PR fixes the problem, or whether it "fixes" a symptom of a different problem, with this code now actively hiding that problem.

teechap:
> This really should have code that reliably reproduces the error, though. 

I agree in principle, but haven't been able to figure out repro steps. According to our bugsnag logs, there isn't a consistent browser with this problem, so there's that. I'm kind of at a loss. It could def be a symptom of a different problem, but do we want this lib to crash in this case rather than, say, log a warning? If so I think we may have to fork this and patch ourselves (no hate against you--thanks for your hard work contributing to open source!).

Pomax:
I don't know who "we" are, but I would absolutely recommend you (temporarily) use your own copy with additional debug code in it until you've found out what your code is doing to cause this, because it sounds very much like there's an interplay either with code your team has written, or code from another library, that's causing you to see this behaviour.

I'm not against putting in a warning, but if you've somehow managed to combine code in such a way that you're trying to access the wrapped instance before it actually got bound (which happens [during the first `render()` pass](https://github.com/Pomax/react-onclickoutside/blob/master/src/index.js#L203-L207), and so cannot be accessed safely until the component has mounted), then that's a strong indicator that your code (or some library's code) is not waiting for the proper time in the React component lifecycle to start doing things.

teechap:
@Pomax thanks for your feedback! I just updated this PR with an `onError` config option and removed the early return in the click handler. Let me know what you think :-)

README.md
```diff
@@ -21,18 +21,19 @@ transpiled es6 classes rather than `createClass` as of v6.
 
 ## Sections covered in this README
 
-* [Installation](#installation)
```
teechap: Sorry about this diff from `*` to `-` in the README. It seems like the pre-commit hook is doing it (I guess I don't have the same version of Prettier?). What do?

Pomax: that's an excellent question. Can you tell it to not kick in for .md files in this specific case?

teechap: To be honest, I don't know how to disable it. Any idea?

teechap: Nevermind, it looks like the `--no-verify` option works...updating

Pomax:
This looks good to me. @Andarist?

vinnymac:
I recently received 13 error reports from users via Sentry on a private repository I work on. The reports are quite detailed, and all 13 reports came from the `return ref.getInstance ? ref.getInstance() : ref;` line.

The error says
```
TypeError
Unable to get property 'getInstance' of undefined or null reference
```

It appears 100% of the users who had this error were using IE 11 (85% on Windows 7, 15% on Windows 10) at the time. Most of my users are using Firefox/Chrome but we do support IE 10 & 11. I have never received an error like this for any other browser and I have been using `react-onclickoutside` since early 2015.

If I can provide any more information, or if you'd like this to be in a separate issue I can move it their, I just thought it related quite well to the name of the PR.

damonbauer:
I'm also getting the `Unable to get property 'getInstance' of undefined or null reference` error. IE & Edge are the only culprits.

Using `react-onclickoutside` version `6.7.1` with React version `16.4.2`

For me, the stack trace starts [here](https://github.com/Pomax/react-onclickoutside/blob/master/dist/react-onclickoutside.js#L271) at `return ref.getInstance ? ref.getInstance() : ref;` then goes back [here](https://github.com/Pomax/react-onclickoutside/blob/master/dist/react-onclickoutside.js#L175) `var instance = _this.getInstance();` and finally back to [here](https://github.com/Pomax/react-onclickoutside/blob/master/dist/react-onclickoutside.js#L207) `if (_this.props.disableOnClickOutside) return;`.

teechap:
FYI, if anyone has concrete steps to repro the problem I would love to see them so I can write test cases for this PR. Just let me know! :-)

vinnymac:
Today I actually got some more reports in with this error. This time from both Edge and IE 11 as @damonbauer says they did above.

The errors from Edge were a little better than those from IE 11, and included breadcrumbs that show me what the user was doing in my application at the time that the error got sent to me. By using them I was able to reproduce the error in a development environment.

Now I have not yet created a reproduction that I can disclose here, but I am one step closer since I can now reproduce this in my closed source project. Once I can abstract some of it into perhaps a codesandbox or something similar I will share it here.

teechap:
@vinnymac thanks for putting this together! Unfortunately I don't have a computer with Windows/Edge, so not sure I can repro without installing a virtual machine etc etc...can anyone else repro? 

FWIW, our bugsnag logs showed this crash on Chrome & Edge, so there are likely multiple possible scenarios where it happens. Hopefully this gets us one step closer to figuring out the issue!

vinnymac:
I modified the reproduction to be a lot simpler, and put the instructions into the codesandbox.io so it is easier to share. If anyone wants to give it a try and needs a VM you can use [this](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/) to download images for Edge on Windows 10. Hopefully someone is able to modify this or make another repro for the issue on Chrome as well.

Here are a list of changes that appeared to prevent the error from being thrown.
1. Replacing `onMouseDown` with `onClick`
2. Replacing `onMouseDown` with `onMouseUp`
3. Changing the `/about/index` Route to `/about`
4. Replacing the Link component with `import { Link } from "react-router-dom";`

It seems very specific to how routing works, whatever the problem is.

Pomax:
If the PR can be updated with a test that actually reproduces the problem, failing on master and passing with this PR, then: absolutely.

artemu78:
> If the PR can be updated with a test that actually reproduces the problem, failing on master and passing with this PR, then: absolutely.

I appreciate your idea.
I`ve tried to make a small example as possible. Also i found react-onclickoutside crushes only being wrapped around third party lib ["react-datetime"](https://www.npmjs.com/package/react-datetime), hope it does not makes you say "its their problems".

Please see small app reproducing crush behavior. 
http://smartoboto.com/issue287/index.html

Please find whole project, with issue attached.
[pack.zip](https://github.com/Pomax/react-onclickoutside/files/2657910/pack.zip)

Pomax:
I appreciate that effort, but I'm afraid I do have to say "if it only happens in combination with one specific other library, and it's not reproducible with just this one, then that's not a PR I can merge in". If you can create a test that can run as part of the tests in https://github.com/Pomax/react-onclickoutside/tree/master/test that run when `npm test` is issued, then that's great, but if there is no way to verify that this (a) fixes things and (b) won't regress then this PR can't really be said to fix a problem, just "change code" =(

Pomax:
That said, ref forwarding is now a thing, so it is probably worth building a new major version that [forwards refs](https://reactjs.org/docs/forwarding-refs.html) and ditches getInstance entirely.

Pomax:
closing this for now, a from-scratch new version based on React 17 with normal modern hooks etc. will be better than trying to revisit this.

PR #288, opened 2018-08-15
fix scroll outside behavior

bogdanyluk:
Currently scrolling outside of the wrapped element does nothing.


Pomax:
Can you explain what this PR addresses in a little more detail, ideally with a link to the issue you opened for it that gives an example of where the current code's doing the wrong thing?

PR #293, opened 2018-09-30
Documented the usage of react-onclickoutside on an SFC

surgeboris:
I've struggled to find "pass handleClickOutside as a prop to wrapped component" documented anywhere, despite of fact that the source code clearly expects that kind of usage: https://github.com/Pomax/react-onclickoutside/blob/master/src/index.js#L73

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ transpiled es6 classes rather than `createClass` as of v6.
 * [Installation](#installation)
 * [Usage:](#usage)
   * [ES6 Class Component](#es6-class-component)
+  * [Handler as `handleClickOutside` prop](#handler-as-handleclickoutside-prop)
+  * [Handler as configuration object property](#handler-as-configuration-object-property)
   * [Functional Component with UseState Hook](#functional-component-with-usestate-hook)
   * [CommonJS Require](#commonjs-require)
 * [Ensuring there's a click handler](#ensuring-there-is-a-click-handler)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,9 +103,25 @@ var clickOutsideConfig = {
...
+
+// create a new component, wrapped below by onClickOutside HOC:
+function MyComponent(props) {
+  // ...
+}
+
+var EnhancedComponent = onClickOutside(MyComponent);
+
+// ...handling code should be passed during EnhancedComponent usage:
+<EnancedComponent handleClickOutside={() => {
```
slurms: _EnhancedComponent_

surgeboris: I've fixed it. Thanks.

README.md
```diff
@@ -79,12 +79,9 @@ var MyComponent = onClickOutside(
 );
 ```
 
-Note that if you try to wrap a React component class without a
-`handleClickOutside(evt)` handler like this, the HOC will throw an error. In
-order to use a custom event handler, you can specify the function to be used by
-the HOC as second parameter (this can be useful in environments like TypeScript,
-where the fact that the wrapped component does not implement the handler can be
-flagged at compile-time):
+Besides having `handleClickOutside(evt)` class method, there are two other ways to pass a custom event handler:
```
Pomax:
```suggestion
In addition to specifying a `handleClickOutside(evt)` class method, there are two alternative ways to pass in a custom event handler:
```

README.md
```diff
@@ -79,12 +79,9 @@ var MyComponent = onClickOutside(
...
+* either specify the function to be used by the HOC in the configuration object (second parameter)
```
Pomax:
```suggestion
* you can specify which function should be used by the HOC, by passing indicating which function should be used in the configuration object, or
```

README.md
```diff
@@ -79,12 +79,9 @@ var MyComponent = onClickOutside(
...
+* or pass function to be used by the HOC as a prop during the usage of the wrapped component
```
Pomax:
```suggestion
* you can pass the function to be used by the HOC as a component property `handleClickOutside={...}` in your JSX
```

README.md
```diff
@@ -106,9 +103,25 @@ var clickOutsideConfig = {
 };
 var EnhancedComponent = onClickOutside(MyComponent, clickOutsideConfig);
 ```
+or:
 
-Note that if you try to wrap a React component with a custom handler that the
-component does not implement, the HOC will throw an error at run-time.
+```js
```
Pomax: I would recommend changing the structure here, too. Rather than stating that there are two ways and then giving a list, I would advocate:
```
There are two ways. The first is to do X:

<<code here>>

The second is Y:

<<code here>>
```

PR #296, opened 2018-10-20
Fixed issue for scenario where handleClickOutside does not get called

jcursoli:
Fixed issue for when manually passing in disableOnClickOutside and preventing handleClickOutside from getting called.


jcursoli:
@Pomax Is there anything I can do to make this PR move a bit faster? Just asking because we use this heavily in our app. Thanks in advance!

Pomax:
If you can add a test that fails without your fix, and passes with your fix, that'll speed things up a lot because it means review goes from "figuring out what is going on", which is a lengthy process, to simply "confirming your test case", which takes almost no time at all.

jcursoli:
@Pomax Okay its been a while but i added the test cases

Pomax:
perfect, I'll have a look at this as soon as I can!

Pomax:
Confirmed that the tests did not work with current, and work with your fix, so: in it goes.

jcursoli:
Thanks a ton!

dist/react-onclickoutside.js
```diff
@@ -293,7 +292,9 @@ function onClickOutsideHOC(WrappedComponent, config) {
         }
       }
 
-      this.componentNode = reactDom.findDOMNode(this.getInstance());
+      this.componentNode = reactDom.findDOMNode(this.getInstance()); // return early so we dont initiate onClickOutside
+
+      if (this.props.disableOnClickOutside) return;
```
tomasztomys: onclickoutside will never initialize if disableOnClickOutside will update (from true to false or from false to true)

tomasztomys: @Pomax

jcursoli: Im trying to understand what you are saying, but isnt that the point of passing disable so it does not get initialised. and if you add it to componentDidUpdate, the handleclickoutside will be initialised.

tomasztomys: If you pass disabled as true and then you change props disabled to false (in component in your app), outside will not be initialized, so will not work..

jcursoli: a quick solution would be to return early in your handclickoutside if your prop is disabled. reverting this commit will make it so its never disabled.

PR #298, opened 2018-11-23
akelon-change

Mayya63:


Pomax:
Which issue does this address? Because:

1. don't comment out code. If it shouldn't be there, remove it.
2. there is zero mention that this fixes an issue already file, so either file an issue and then say this fixes it, or this PR is getting closed.

Pomax:
closing due to no response.

PR #301, opened 2018-11-24
Use forwardRef

### LINKED ISSUE #297: Using react-reactonlickoutside with React.forwardRef results in a runtime error
Using the following component that uses [React.forwardRef](https://reactjs.org/docs/forwarding-refs.html):
```JSX
const TestComponent = onClickOutside(
    React.forwardRef((props, ref) => <div {...props} ref={ref} />)
);
```
Leads to the following error:
```
Uncaught TypeError: Cannot read property 'isReactComponent' of undefined
    at onClickOutside.render (react-onclickoutside.es.js:329)
    at finishClassComponent (react-dom.development.js:13727)
    at updateClassComponent (react-dom.development.js:13690)
    at beginWork (react-dom.development.js:14489)
    at performUnitOfWork (react-dom.development.js:17014)
    at workLoop (react-dom.development.js:17054)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
    at replayUnitOfWork (react-dom.development.js:16366)
```
As `styled-components` uses `React.forwardRef`, 'react-onclickoutside' cannot be directly used with styled components.

Andarist:
fixes #297 

this doesn't work yet - need to decide what we want to do about function components (they HAVE to be wrapped with `forwardRef` now - wrapped component cant be "just" a function component)

also IMHO ref should be treated as DOM component now - currently the implementation allows it to be a composite component and handles its class methods, this should be IMHO removed - the handler has to be provided in config or in props, not as property of the wrapped component

and we should completely remove `findDOMNode` usage

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,8 +66,8 @@
     "mocha": "^4.0.1",
     "npm-run-all": "^4.0.2",
     "prettier": "^1.8.2",
-    "react": "^15.5.x",
-    "react-dom": "^15.5.x",
+    "react": "^16.6.3",
+    "react-dom": "^16.6.3",
     "react-test-renderer": "^15.5.x",
     "require-hijack": "^1.2.1",
     "rimraf": "^2.6.2",
@@ -77,8 +77,8 @@
     "webpack": "^3.8.1"
   },
   "peerDependencies": {
-    "react": "^15.5.x || ^16.x",
-    "react-dom": "^15.5.x || ^16.x"
+    "react": "^16.3.0",
+    "react-dom": "^16.3.0"
   },
   "lint-staged": {
     "{src,test}/**/*.js": [
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,4 @@
-import { createElement, Component } from 'react';
+import { createElement, Component, forwardRef } from 'react';
 import { findDOMNode } from 'react-dom';
 import * as DOMHelpers from './dom-helpers';
 import { testPassiveEventSupport } from './detect-passive-events';
@@ -33,7 +33,7 @@ function getEventHandlerOptions(instance, eventName) {
  * onClickOutsideHOC function defined inside setupHOC().
  */
 export default function onClickOutsideHOC(WrappedComponent, config) {
-  return class onClickOutside extends Component {
+  class OnClickOutside extends Component {
     static displayName = `OnClickOutside(${WrappedComponent.displayName || WrappedComponent.name || 'Component'})`;
 
     static defaultProps = {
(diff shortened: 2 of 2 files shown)


Pomax:
Yeah I think if we're doing a new major, totally worth doing. Change the requirement to say "please pass in a callback" and ditch the attemp at forwarding the instance(s)/nodes

npeterkamps:
@Andarist Any progress on this? This is currently blocking the migration to styled-components v4, so it would be great if this gets fixed.

Andarist:
Truth to be told - I've totally forgot that I've started working on this. Would u be willing to continue the work here? Not sure when I am gonna get some free time to finish this.

npeterkamps:
Sure. I'd be happy to help out.

> also IMHO ref should be treated as DOM component now - currently the implementation allows it to be a composite component and handles its class methods, this should be IMHO removed - the handler has to be provided in config or in props, not as property of the wrapped component

If I understand correctly, we still need to obtain the instance for the config callback? So we either use the `onClickOutsideHOC`'s `config.handleClickOutside(instance)` or `this.props` to determine the `handleClickOutside`?

> and we should completely remove `findDOMNode` usage

Does something replace `findDOMNode`? `componentNode` is used for the ignore behavior, so how would that work if we don't use `findDOMNode`?

Are there any other issues that should be fixed before the next major release? Maybe we could create a milestone?

Any plans to change the API of the HOC itself? Most HOCs are a function that takes config, which returns a function that takes the wrapped component, which returns the HOC, e.g. `export default onClickOutside({ ...config })(WrappedComponent)`.

Pomax:
That pattern is for getting the hoc to generate a function, which is then used to wrap the component, which feels an unnecessary step of indirection here. Having the `onClickOutside()` function return a component, by passing in a component and (entirely) optional config object as second argument feels just fine.

Andarist:
> Does something replace findDOMNode? componentNode is used for the ignore behavior, so how would that work if we don't use findDOMNode?

We still need access to `componentNode` - we can just make it user's responsibility to "connect" pieces together. 

After thinking about this some more right now it won't be as clean as it could be though, because it forces the consumer to use `React.forwardRef` too to name the `ref` however they want and pass received (from onClickOutside) ref callback to the underlaying DOM element. The manual plumbing is fine for me, bummer is the required `React.forwardRef` 🤔 WDYT about this @Pomax ? Alternative would be to continue passing `innerRef` prop to the wrapped component (with an option to configure the name maybe). 

When hook came out I've experimented with them a little bit and I've tried reimplementing this HOC as hook ( https://github.com/Andarist/use-onclickoutside ) and the implementation there is much cleaner in regards to this manual plumbing, I just require `ref` from the consumer (requiring a ref rather returning one increases composability quite a lot). We ain't living in hooks world yet though, so we have to figure out best HOC API right now

arthur-cen:
Any updates on this feature?

Pomax:
@Andarist should this PR be closed as abandoned? Or is it very little work to put the finishing touches on?

Andarist:
I think it wouldn't take much to finish this - you'd only have to resolve concerns mentioned in the thread here. I'm afraid I won't have time to do this myself though, also as the PR itself is not a long one it might be easier to start from scratch.

Pomax:
closing this for now, a from-scratch new version based on React 17 with normal modern hooks etc. will be better than trying to revisit this.

PR #308, opened 2019-01-24
Add | Code Example in the Readme for usage with Functional Components …

ZeroPie:
Add | Code Example in the Readme for usage with Functional Components and Hooks
Add | Section Title for Usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ transpiled es6 classes rather than `createClass` as of v6.
 ## Sections covered in this README
 
 * [Installation](#installation)
+* [Usage](#usage)
 * [Regulate which events to listen for](#regulate-which-events-to-listen-for)
 * [Regulate whether or not to listen for outside clicks](#regulate-whether-or-not-to-listen-for-outside-clicks)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +22,7 @@ transpiled es6 classes rather than `createClass` as of v6.
 ## Sections covered in this README
 
 * [Installation](#installation)
+* [Usage](#usage)
```
Pomax: we might as well also get those nested entries in, as it'll let people laser-focus on what they're looking for. Probably indented?

README.md
```diff
@@ -79,6 +108,8 @@ var MyComponent = onClickOutside(
 );
 ```
 
+---
```
Pomax: horizontal rules usually mean there's a missing heading, let's replace that `---` with a heading `### ensuring there's a click handler"?

README.md
ZeroPie: Good idea, usage headings are now indented under Usage)

Pomax:
sweet, thanks for the help!

Pomax:
how about I also merge it. yes, great idea.

PR #311, opened 2019-02-19
prevents calling findDOMNode on unmounted instance

capaj:
react throws error such as this when findDOMNode is used on an unmounted component instance.
For me this happens when using react-datetime. When user clicks outside the state of our app is mutated and we rerender the view containing the react-datetime and react-onclickoutside components.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -114,7 +114,9 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
     }
 
     componentDidUpdate() {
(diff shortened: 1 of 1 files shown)


Pomax:
Hm, I'm a little confused as to how an unmounted component could receive updates. What do those updates even do if there's no associated component to show the updates for? (e.g. that sounds like accidentally using a component as if it's a normal JS object instead of being part of a user interface)

PR #322, opened 2019-05-20
Fixed one popover in another in IE11 after unsubscribing events

savtym:
If we have one popover in another in IE11, after unsubscribing events IE calls callback for nested popover while this popover was unsubscribed.

So, I have an error `Unable to get property 'getInstance' of undefined or null reference` for this method `_this.getInstance`.

I propose to add variable `_isSubscribeEvents` for class `onClickOutside` which will define subsribe/unsubscribe events in browsers.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -64,6 +64,12 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
     }
 
     __outsideClickHandler = event => {
+      // If we have one popover in another in IE11, after unsubscribing events
+      // IE calls callback for nested popover while this popover was unsubscribed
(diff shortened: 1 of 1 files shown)


Pomax:
Are you using separate configurations for each menu?

Pomax:
Also I don't want to be landing code that specifically fixes something only for IE11, which is the last flavour of a browser not even Microsoft wants people to use anymore, so I'd first like to know what the exact situation is here, and whether there is a solution in terms of "how this HOC got used". There should be any cross communication between two separately wrapped components, as they get locally bound functionality, so I'd like to ask if you can show code that demonstrates the problem (as a minimal example of course)

PR #323, opened 2019-05-23
make it work with shadow dom

cameronbraid:
This seems to fix shadow dom issues.

The basic issue is that with a dom structure like :

- document
  - body
    - div#shadow-host
      - #shadow-root
         - div#wrapper
           - onClickOutsideDecoratedComponent
             - div#clicker

In the above case the mousedown listener is registered on `document` and when clicking on `div#clicker` the handler is called with an event with target = `div#shadow-host` where as we really want it to start at `div#wrapper`

So the solution is to add the event listeners to both `document` and `#shadow-root` 

However this causes the handler is triggered twice, once with event.target = `div#clicker`, then once with event.target=`div#shadow-host`.  So we tag the event and only handle it once, from `div#clicker`

Then when walking up the dom in findHighest we have to continue via the host element so that we traverse the host document as well

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -12,6 +12,49 @@ const enabledInstances = {};
 const touchEvents = ['touchstart', 'touchmove'];
 export const IGNORE_CLASS_NAME = 'ignore-react-onclickoutside';
 
+function getDocumentNodes(node) {
+
+  let docs = []
+  while (node.parentNode) {
+    node = node.parentNode
+    
+    if (node.host) {
+      docs.push(node)
+      node = node.host
+    }
+
+  }
+  docs.push(document)
+  return docs
+}
+
+/**
+ * Try to find our node in a hierarchy of nodes, returning the document
+ * node as highest node if our node is not found in the path up.
+ */
+function findHighest(current, componentNode, ignoreClass) {
+  if (current === componentNode) {
+    return true;
+  }
+  // If source=local then this event came from 'somewhere'
+  // inside and should be ignored. We could handle this with
+  // a layered approach, too, but that requires going back to
+  // thinking in terms of Dom node nesting, running counter
+  // to React's 'you shouldn't care about the DOM' philosophy.
+
+
+  while (current.parentNode) {
+    if (isNodeFound(current, componentNode, ignoreClass)) {
+      return true;
+    }
+
+    // walk up to the parentNode, or if we are at a shadow root, continue up via the host element
+    current = current.parentNode || current.host
+  }
+
+  return current;
+}
+
 /**
  * Options for addEventHandler and removeEventHandler
  */
@@ -149,6 +192,9 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
       handlersMap[this._uid] = event => {
         if (this.componentNode === null) return;
 
+          if (event.PROCESSED) return
+          event.PROCESSED = true
+
         if (this.props.preventDefault) {
           event.preventDefault();
         }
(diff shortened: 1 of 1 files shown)


cameronbraid:
I've opened this as a draft PR since I really just wanted to get a discussion going on the correct solution for this

Pomax:
fair enough - I'm not sure this is a solvable issue in this way since the shadow dom is supposed to be genuinely "hidden" from the page, so as far as the browser can tell the page, there's nothing "in" the shadow DOM, the event simply started in the (custom) element that is found in the regular DOM. 

What's the concrete use case we're trying to solve for here, though? Because React itself doesn't use shadow DOM, that's more a custom elements thing (which I would love to see get so easy to use they basically make React obsolete =D).

cameronbraid:
My primary case is css isolation.

By rendering into a shadow DOM I am albe to use my component in contexts where I have no control over the styling of the page.  For example if I use semantic ui and a third party use bootstrap there will be no css clashes when embedding my control

I am using a component (react-datepicker) that uses onclickoutside to hide the calendar popup when you click outside the popup.  However when used within a shadow dom, clicks within the calendar popup always close the popup due to the above mentioned reason.

newying61:
@cameronbraid 
Got the same issue in shadowDOM.
Another solution maybe checking the event.composed flag.
If it's true, then call the event.composedPath(), inside the returned array the first one is the real target.
In older browsers, if we use the webcomponentsjs polyfill, shadowRoot is a virtual element in memory,
it seems addEventListener is not working sometimes, haven't chase why.

Pomax:
Looking at [Event.composed](https://developer.mozilla.org/en-US/docs/Web/API/Event/composed) is a good call, because it lets us access [Event.composedPath](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath), which is probably the exact thing we want to be looking at in this case, rather than having to do a "while not at top" loop.

newying61:
@Pomax @cameronbraid 
I've raised another pull request #324 .
Please have a look.

The polyfill is for custom web components API. Not for this lib. This lib works quite well in Chrome and other modern browsers. :)

mbIkola:
Faced same issue when using react-datepicker in shadowRoot.  Patch applied as "temporary workaround":
```diff
diff --git a/src/index.js b/src/index.js
index 3e66bf8..3ec0e85 100644
--- a/src/index.js
+++ b/src/index.js
@@ -131,6 +131,8 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
      * for clicks and touches outside of this element.
      */
     enableOnClickOutside = () => {
+      const document = this.componentNode && this.componentNode.getRootNode ? this.componentNode.getRootNode() : document;
+
       if (typeof document === 'undefined' || enabledInstances[this._uid]) {
         return;
       }
@@ -178,6 +180,8 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
      * for clicks and touches outside of this element.
      */
     disableOnClickOutside = () => {
+      const document = this.componentNode && this.componentNode.getRootNode ? this.componentNode.getRootNode() : document;
+
       delete enabledInstances[this._uid];
       const fn = handlersMap[this._uid];

```
quite similar to solution in this PR, but a bit simpler :) 
I can't use composedPath, cuz in my env shadowTree is attached in closed mode.

Pomax:
closing due to inactivity (mine, mostly, and unless someone is going to fund me working on this again, it's unlikely that'll change)

Issue #325, opened 2019-06-21
OCO applies only to last functional component


EdPutans:
When multiple functional components are rendered by mapping, the onclickoutside only applies to the last component. This only occurs when defining the component as a stateless function, but works just fine when it's a class component.

Pomax:
This feels like a consquence of how this HOC works - if you need more than one of these on a page, you're probably not better off with a functional component, and that thing should probably be a class. See https://github.com/Pomax/react-onclickoutside/issues/320 and https://github.com/Pomax/react-onclickoutside/issues/307

jRichardeau:
Can you update the readme with this ? I lost a couple of hours with it.

PR #326, opened 2019-06-30
Support forward ref over hooks

### LINKED ISSUE #320: How to use two or more  instance?
How can are use two or more  instance component? If you use the component more than once, then the outsider click does not work.  Only on the last component. And it works weird. If this is not a bug, then is it possible to solve the problem? Thank.

example problems:
 https://codesandbox.io/s/yw45z06749?fontsize=14

NilSet:
Fixes #320 #325 #297

using forwardRef and useImperativeHandle you can now return a callback
properly bound to a particular element.


Pomax:
There's a whole bunch of code in the `dist` file that cannot be explained merely by your changes to the `src/index` file... would you happen to know where that code came from?

src/index.js
```diff
@@ -25,6 +25,20 @@ function getEventHandlerOptions(instance, eventName) {
   return handlerOptions;
 }
 
+// borrowed from react-is
+const hasSymbol = typeof Symbol === 'function' && Symbol.for;
+const REACT_FORWARD_REF_TYPE = hasSymbol ? Symbol.for('react.forward_ref') : 0xead0;
```
Pomax: This is not necessary. Every in-support browser [supports `Symbol`](https://caniuse.com/#search=Symbol). People who need to support dead browsers are responsible for polyfilling their own site.

src/index.js
```diff
@@ -25,6 +25,20 @@ function getEventHandlerOptions(instance, eventName) {
...
+
+/**
+ * Check if component is either a forwardRef or a class component
+ */
+function acceptsRef(WrappedComponent) {
+  return (
+    (WrappedComponent.$$typeof && WrappedComponent.$$typeof === REACT_FORWARD_REF_TYPE) ||
```
Pomax: I am unclear where this `$$type` is coming from, but I'd very much like to not use `$` in var names when the code already uses `_` to mark private functions and variables.

src/index.js
NilSet: I was attempting to duplicate the relevant code from `react-is`. I would have just pulled in the package as a dependency or peer dependency but I had difficulty figuring out your build system.

If you like I can simplify and remove the checks, and assume that symbols were available when react was loaded, but this is a one time check at load time.

src/index.js
NilSet: `$$typeof` is an internal react property used to brand react elements and components. It is used by `react-is`, as well as by react internals.

NilSet:
Regarding the changes in dist, it is because I upgraded the suite of babel dependencies to what was at the time the latest version. They had been stuck on an old beta build before this.

src/index.js
Pomax: `Symbol` is a JavaScript language feature, not a React feature, so there's no need to perform this check: we can trust they exist.

src/index.js
Pomax: Is this publically documented? If not, let's not use it, because it cannot be relied on unless there's official documentation that says that's what it's called, and that's what it's for.

Pomax:
Hm, I see. The build system may need some updated rules, then, too. It seems to be generating code that is entirely unnecessary.

src/index.js
Pomax: Reading through https://reactjs.org/docs/forwarding-refs.html, it looks like the actual way to check is to, at run time, see if there's a `ref.current`. If there is, we're dealing with a forwarded ref. If there is not, we're dealing directly with a component.

src/index.js
NilSet: To clarify, React itself uses different brands depending on if Symbol exists. The code to check the brand depends on whether Symbol existed at the time React was loaded. If you don't want to support pre-symbol browsers, I can remove the  magic number and always assume the brand is a Symbol.

src/index.js
NilSet: It is not publicly documented, but it is used in the separately published first party package react-is https://github.com/facebook/react/blob/master/packages/react-is

It is not an artifact of the build system, but it appears in the source code of react-is.
https://github.com/facebook/react/blob/master/packages/react-is/src/ReactIs.js#L30
https://github.com/facebook/react/blob/master/packages/shared/ReactSymbols.js

Also, react-is does not have any version constraints on what versions of react it is compatible with, which I can only assume means the brands are stable. https://github.com/facebook/react/blob/master/packages/react-is/package.json

NilSet:
Regarding `$$typeof`, it would not be necessary to use the internals if we instead depend on `react-is`, presumably as a peer dependency, possibly with a fallback for when it cannot be found.

On the other hand, this package is already using `isReactComponent` which afaict is also an undocumented internal property.

NilSet:
> Hm, I see. The build system may need some updated rules, then, too. It seems to be generating code that is entirely unnecessary.

The only net new code I see is `_assertThisInitialized`, is this what you were talking about?

npeterkamps:
@Pomax Could you please take a look at this pull request? Would be really nice if this could get merged.

Djiit:
Hi there, we could really benefit from this PR ;)

Pomax:
closing due to inactivity - mine, mostly, and unless someone is willing to fund me working on this again, that's unlikely to change.

PR #334, opened 2019-10-28
Fix to accommodate arrow functions and their lack of a prototype.

justnoah:
While trying to implement testing, I was getting the error:
  - _Cannot read property 'isReactComponent' of undefined_

After some diligent internet scouring, I came across a similar issue on **react-table**, where they came to the conclusion that the lack of a prototype on arrow functions was the root cause and the method that they had in place to normalize, didn't cover that.  @rpalermodrums suggested using _Object.getPrototypeOf_ as a solution. It works for me locally, so fingers crossed. I'm still a little new to software engineering, so I can't say that I'm fully confident that I understand what the potential ramifications are, although it seems pretty straight-forward. Please don't kill me.

### DIFF
--- a/dist/react-onclickoutside.js
+++ b/dist/react-onclickoutside.js
@@ -277,7 +277,7 @@ function onClickOutsideHOC(WrappedComponent, config) {
     var _proto = onClickOutside.prototype;
 
     _proto.getInstance = function getInstance() {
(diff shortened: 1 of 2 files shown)


sedlukha:
Can this be merged?

DiFuks:
Why was it closed? :( I have the same problem with arrow function components

Issue #346, opened 2021-01-25
Support React 17.x


h3rmanj:
Can't install into projects with npm@7 and react@17 at the moment.
[React 17](https://reactjs.org/blog/2020/10/20/react-v17.html) doesn't really add anything, so it shouldn't break stuff.

PR #347, opened 2021-01-25
Update react & react-dom peer-dep to support 17.x

h3rmanj:
Addresses #346

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,8 +80,8 @@
     "webpack": "^3.8.1"
   },
   "peerDependencies": {
-    "react": "^15.5.x || ^16.x",
-    "react-dom": "^15.5.x || ^16.x"
+    "react": "^15.5.x || ^16.x || ^17.x",
+    "react-dom": "^15.5.x || ^16.x || ^17.x"
   },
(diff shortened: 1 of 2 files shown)


Pomax:
Unfortunately, this alone is not enough to make things work. The CI task is currently failing on uglify, but I'm sure there are more deps that need to be checked before this will pass.

Pomax (changes requested):
CI does not pass with just these updates.

h3rmanj:
The latest release 6.7.1 failed for the same reason, i'm guessing there is some old build pipeline here?
This PR is only to fix the issue of not being able to install this package at all if you have npm v7, and react v17.

h3rmanj:
The `warnings` flag is a top-level flag to uglify, so I moved it out of the `compress` options. Seems to have fixed the pipeline.
It's [`false` by default](https://github.com/mishoo/UglifyJS#minify-options) though, but i left it there for good measure.

Pomax:
cheers. I'll merge this in and push a minor update.

Pomax:
v6.10.0 published

PR #356, opened 2021-05-11
Use empty object instead of null as default handlerOptions

egudkov:
Hi!

## Motivation
Our team found a conflict between the code of `react-onclickoutside` and the code of some wordpress plugins and themes.
Those plugins use code that makes event listeners passive by default in order to improve scrolling experience, see example by the link - https://stackoverflow.com/a/45974787

This code overwrites `addEventListener` and according to the following lines...
```js
    EventTarget.prototype.addEventListener = function(type, listener, options) {
      var usesListenerOptions = typeof options === 'object';
      var useCapture = usesListenerOptions ? options.capture : options;
```
...there will be an error if `null` is passed as `options`, because `typeof null === 'object'` is true and then `null.capture` leads to a `TypeError: Cannot read property 'capture' of null` error.

And currently there is the case when `react-onclickoutside` passes `null` as `options` to `addEventListener`.
So it leads to the error when our product (with `react-onclickoutside`) is embedded to the user site that has the wordpress plugin.

## Changes
The proposal is to use empty object as default options value instead of null.
It looks like more expected behaviour [according to specs](https://dom.spec.whatwg.org/#dom-eventtarget-addeventlistener-type-callback-options-options) and it fixes the conflict.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -16,12 +16,13 @@ export const IGNORE_CLASS_NAME = 'ignore-react-onclickoutside';
  * Options for addEventHandler and removeEventHandler
  */
 function getEventHandlerOptions(instance, eventName) {
-  let handlerOptions = null;
+  const handlerOptions = {};
   const isTouchEvent = touchEvents.indexOf(eventName) !== -1;
 
   if (isTouchEvent && passiveEventSupport) {
-    handlerOptions = { passive: !instance.props.preventDefault };
+    handlerOptions.passive = !instance.props.preventDefault;
   }
+
   return handlerOptions;
 }
 


Pomax:
I'd like to point out that overriding the `addEventListener` function in the way you're showing is pretty much its own bug right there, and should be reported and fixed. Adding a `null` test to that code should be trivial and fix things for _every_ library.

egudkov:
I agree 100% - there should be check for null. The problem is that there are at least two plugins with this bug, maybe more. I will contact the authors and ask to add a `null` test but I'm not sure that the bug will not come back later with other plugins. So, the idea is to also make a change in this lib to prevent new cases of that bug.

Anyway, we are going to delete the component where we used `react-onclickoutside` lib (not because of the described problem) and this case will not be an issue for us. So, I guess it is up to you to decide what to do with this pull request.

Pomax:
Cheers. If you can drop a link to the issue you filed with them once you file it, I will probably merge this in. (but not before, because the most important thing is that they fix it. Until they do, me fixing this lib's code just encourages them to not bother fixing their code).

Pomax (approved):

PR #375, opened 2022-01-31
Removed issue with findDOMNode

MaksZhukov:
Removed issue with findDOMNode


README.md
```diff
@@ -88,13 +71,13 @@ function listenForOutsideClicks(listening, setListening, menuRef, setIsOpen) {
     if (listening) return;
     if (!menuRef.current) return;
     setListening(true);
-    [`click`, `touchstart`].forEach((type) => {
-      document.addEventListener(`click`, (evt) => {
+    [`click`, `touchstart`].forEach(type => {
```
Pomax, line 74: following standard `prettier` formatting, no exception to parentheses. None, one, or more, all use parens.

README.md
```diff
@@ -430,3 +407,7 @@ for, then the README is missing important information that should be added in.
 Please [file an issue](https://github.com/Pomax/react-onclickoutside/issues) with a request for additional documentation,
 describing what you were hoping to find in enough detail that it can be used to
 write up the information you needed.
+
+## In addition
+
+To avoid findDOMNode use "outsideNodeRef" props from a wrapping component and set it to element as ref.
```
Pomax, line 413: People's own code doesn't use `findDOMNode`, can you explain why this section would be needed?

Pomax (changes requested):
thanks fo the PR but let's not apply autoformatting to the readme, it looks like it mangled the text and did some replacements that weren't necessary.

README.md
MaksZhukov: Developers should set outsideNodeRef themselves to avoid `findDOMNode` issue

Pomax: then let's move this a whole lot up, because that makes this an important note that should be part of the "how to use" instructions, rather than sneaking it in at the very end of the readme.

We should a slighlty longer explanation and then I'd say let's put this directly below the "Ensuring there is a click handler" section, with a code example, and an explanation of why people want to do this, with a directly quoted error so that if folks run into that error, and search the readme, they find the section that explains how to fix it.

README.md
```diff
@@ -31,23 +30,27 @@ transpiled es6 classes rather than `createClass` as of v6.
...
-* [Regulate whether or not to listen to scrollbar clicks](#regulate-whether-or-not-to-listen-to-scrollbar-clicks)
-* [Regulating `evt.preventDefault()` and `evt.stopPropagation()`](#regulating-evtpreventdefault-and-evtstoppropagation)
-* [Marking elements as "skip over this one" during the event loop](#marking-elements-as-skip-over-this-one-during-the-event-loop)
-* [Older React code: "What happened to the Mixin??"](#older-react-code-what-happened-to-the-mixin)
-  * [But how can I access my component? It has an API that I rely on!](#but-how-can-i-access-my-component-it-has-an-api-that-i-rely-on)
-* [Which version do I need for which version of React?](#which-version-do-i-need-for-which-version-of-react)
-  * [Support-wise, only the latest version will receive updates and bug fixes.](#support-wise-only-the-latest-version-will-receive-updates-and-bug-fixes)
-* [IE does not support classList for SVG elements!](#ie-does-not-support-classlist-for-svg-elements)
-* [I can't find what I need in the README](#i-cant-find-what-i-need-in-the-readme)
+* [:warning: Open source is free, but developer time isn't :warning:](#warning-open-source-is-free-but-developer-time-isnt-warning)
```
Pomax, line 33: Please restore the original ToC. A ToC is for indexing main content, not "every heading in the document". Anything preceding the ToC does not go in it.

README.md
Pomax: please still restore this to its previous form.

Pomax (changes requested):

MaksZhukov:
@Pomax

Pomax:
Unfortunately I'm currently dealing with an arm injury, so this will have to wait a while until that's healed up enough that I can do FOSS work in my spare time.

MaksZhukov:
@Pomax Hey! I hope you are well.

Issue #395, opened 2023-04-20
Compatibility Issues Between "React" and "React-OnClickOutside" Libraries.


OrMeirchak:
Our project currently has 'react-onclickoutside': '^6.7.0' installed.
 We recently updated our project from 'react': '^16.2.0' to 'react': '^18.2.0'.
 However, we have encountered an issue with a component that is wrapped in onClickOutside.
 After upgrading to the new version of React, the handleClickOutside function is called immediately after the component is created, without the user performing any action. 
When we rolled back to 'react': '^16.2.0', the problem was resolved.

We would like to know if there is a version of 'react-onclickoutside' that is supported by 'react': '^18.2.0'?

Pomax:
I assume you kept the class notation rather than rewriting things to functions with hooks, like the React folks have been pushing for since 16.5?

OrMeirchak:
> 

Hi Pomax,

Thank you for your reply.

I am working on a large project that has been in development for many years. As a result, a significant portion of our components are still written using class notation.

Pomax:
Just making sure. I don't suppose you have a small self-contained example that can be dropped into the issue for testing?

OrMeirchak:
Hi pomax , thank you very much for your willingness to help.

Unfortunately, I was not able to create an example that I could share.
 After a long time of debugging, I noticed that during the creation of the wrapped component, it is registered to the event in this line:
`document.addEventListener(eventName, handlersMap[this._uid], getEventHandlerOptions(this, eventName));`
And immediately after that, the _**handlersMap**_ function is activated. (The wrapped component is created by clicking on the screen that creates the event)

This is a behavior that also occurs in the old version of React, but for some reason, the event does not happen for the component that was just registered.

Thanks,or

Pomax:
Remember that an example doesn't need to be "your code". If you're seeing it do the wrong thing, it should be possible to create a standalone "new code" setup with a few dummy components that shows things going wrong. (because if you can't, there aren't really any steps to reproduce the problem, which means we can't make a test case)

OrMeirchak:
Hi Pomax,
 I've found a solution to the problem. In the constructor of the wrapped component, I save an initTimeStamp. 
Then, every time before I execute the logic of the handleClickOutside function in the wrapped component, I check if the timeStamp of the event is greater than the initTimeStamp.
 Do you now understand better what is causing the problem, or would you like me to try to create a pull request?"

OrMeirchak:
[...]

PR #396, opened 2023-05-20
Compatibility Issues Between "React" and "React-OnClickOutside" Libraries

OrMeirchak:
Fix for the problem where the event handler was being called even though it was registered after the event was created
[](
https://github.com/Pomax/react-onclickoutside/issues/395)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -51,6 +51,7 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
     constructor(props) {
       super(props);
       this._uid = uid();
+      this.initTimeStamp = performance.now();
     }
 
     /**
@@ -163,6 +164,8 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
 
       handlersMap[this._uid] = event => {
         if (this.componentNode === null) return;
+        
+        if (this.initTimeStamp > event.timeStamp) return;
 
         if (this.props.preventDefault) {
           event.preventDefault();


src/index.js
```diff
@@ -163,6 +164,8 @@ export default function onClickOutsideHOC(WrappedComponent, config) {
 
       handlersMap[this._uid] = event => {
         if (this.componentNode === null) return;
+        
+        if (this.initTimeStamp > event.timeStamp) return;
```
Pomax, line 168: Hmm, this feels like it's fixing a race condition. This function should not be able to trigger before the constructor runs, so while this might fix the symptom it's probably better to track down the underlying cause. Thanks to it being a long weekend here in Canada, I should hopefully have time today to dig into this a little.

akmittal: Was there any better solution discussed for this?

Pomax: I guess the counter question would be that modern React with hooks kind of... [doesn't need this hoc anymore](https://github.com/Pomax/react-onclickoutside?tab=readme-ov-file#functional-component-with-usestate-hook), does anyone still _need_ this fixed?

akmittal: We are working on an oldcodebase where its not feasible to migrate to hooks easily, So it would be great if we can fix a solution for the issue.

Pomax (approved):

src/index.js
Pomax: Hooks are over 5 years old at this point, I would _strongly_ encourage you to consider getting on board with modern Reactg if you want to keep using it (understanding full well that's almost certainly not your call to make).

I'll merge this in, but given React's trajectory, this will be a bandaid at best.

Pomax: Also, if this HoC is _that_ important to your codebase, consider taking the first paragraph of the README to heart and giving back any way you can.

Pomax:
published v6.13.1

